PR #14, opened 2024-10-15
Playground

brenkao:
- Added Playground - Click Open Playground in Traces Table.
    - You can create new versions
    - You can vibe check, which will also create new version if any changes are made
- Added Components for Call Params 
    - Added only mandatory call params
    - Optional call params have not been added yet
- Made major changes to db, this is a breaking change. Hopefully last breaking change.
- Fixed various UI bugs when viewing traces
- Added more error checking when writing prompts


lilypad/server/main.py
```diff
@@ -208,6 +210,113 @@ async def get_llm_function_id_by_hash(
...
 
 
+def create_dynamic_function(
+    name: str, args_dict: dict[str, str], return_annotation: type
+):
+    """Create a dynamic function with the given name, arguments, and return type."""
+    arg_list = [f"{arg_name}: {arg_type}" for arg_name, arg_type in args_dict.items()]
+    arg_string = ", ".join(arg_list)
+
+    func_def = f"def {name}({arg_string}) -> {return_annotation.__name__}:\n    ..."
```
willbakst, line 220: nit: does this need the `\n`?

brenkao: oh, no it doesnt

lilypad/server/main.py
```diff
@@ -206,6 +198,84 @@ async def get_llm_function_id_by_hash(
...
+    )
+
+    fn = create_dynamic_function(
+        name=version.llm_fn.function_name,
+        args_dict=args_dict,
+        return_annotation=str,
+    )
+
+    if not version.fn_params:
+        return trace_decorator(fn)(**arg_values)
```
brenkao, line 251: should i just type ignore this or is there a way to get typing

lilypad/server/main.py
```diff
@@ -206,6 +198,84 @@ async def get_llm_function_id_by_hash(
...
+
+    fn_params = FnParamsPublic.model_validate(version.fn_params)
+    return traced_synced_llm_function_constructor(fn_params, trace_decorator)(fn)(
```
brenkao, line 254: ditto

lilypad/models/fn_params.py
```diff
@@ -1,24 +1,63 @@
...
+class ResponseFormat(BaseModel):
+    """Response format model."""
+
+    type: Literal["text", "json_object", "json_schema"]
+
+
+class OpenAICallArgsCreate(BaseModel):
+    """OpenAI call args model."""
+
+    max_tokens: int
```
willbakst, line 19: not all of these are required fields?

lilypad/models/fn_params.py
```diff
@@ -1,24 +1,63 @@
...
+    frequency_penalty: float
+    presence_penalty: float
+    response_format: ResponseFormat
+    stop: str | list[str] | None = None
+
+
+class AnthropicCallArgsCreate(BaseModel):
+    """Anthropic call args model."""
+
+    max_tokens: int
```
willbakst, line 31: same here

lilypad/cli/commands/start.py
```diff
@@ -51,7 +51,7 @@ def start_lilypad(project_dir: Path, port: str) -> subprocess.Popen:
     env["LILYPAD_PROJECT_DIR"] = str(project_dir)
     files_dir = files("lilypad").joinpath("server")
     process = subprocess.Popen(
-        ["fastapi", "dev", "--port", port],
+        ["fastapi", "dev", "--port", str(port)],
```
willbakst, line 54: there's another issue with `lilypad.configure` where it hardcodes port 8000 and instead needs to read from the config file, please fix that

lilypad/models/fn_params.py
brenkao: they arent required but they have default values, so they are always set to default in the UI

lilypad/server/main.py
willbakst: just use the specific `pyright: ignore [{SPECIFIC_ERROR}]` for now, we can go through type ignores later and fix them, but not super high priority right now so long as it works

lilypad/server/main.py
willbakst: same here as my other comment

lilypad/cli/commands/start.py
brenkao:
```
    config = load_config()
    port = config.get("port", 8000)
    otlp_exporter = JSONSpanExporter(
        base_url=f"http://127.0.0.1:{port}/api",
    )
```
this is the code inside configure, are you saying its not pulling the port properly?

willbakst: oh hmm, for some reason it wasn't tracing, not sure why

brenkao: there was a bug in start that didnt set the config file properly, which may have caused a weird state.

lilypad/models/fn_params.py
willbakst: but many of them can be null/none? the api doesn't require them

lilypad/models/fn_params.py
willbakst: max_tokens is required, but not all others are

willbakst (approved):

PR #17, opened 2024-10-25
Development lifecycle updates

brenkao:
* Refactored majority of SQL statements into Services for easier testing in the future
* Updated definition of version for synced prompts. All version checking checks function hash instead of function name.
* Removed edit CLI command until we figure out if we want to keep that flow
* Added Mirascope traces via synced and unsynced + mirascope
* Added BaseModel support for unsynced
* Tools, BaseModel, show up but are not pretty
* Added active button and code for better UX in the playground
* Added UI cards for OpenAI, Gemini, and Anthropic
* No support for Gemini for synced

### DIFF
--- a/lilypad/_utils.py
+++ b/lilypad/_utils.py
@@ -5,14 +5,17 @@ import json
 import os
 import time
 import webbrowser
-from collections.abc import AsyncIterable, Callable, Coroutine, Iterable
+from collections.abc import AsyncIterable, Callable, Coroutine, Generator, Iterable
+from contextlib import _GeneratorContextManager, contextmanager
 from functools import partial, wraps
 from importlib import import_module
 from pathlib import Path
 from typing import Any, ParamSpec, TypeVar, cast, get_args, get_origin, get_type_hints
 
 from mirascope.core import base as mb
+from opentelemetry.trace import get_tracer
 from opentelemetry.trace.span import Span
+from opentelemetry.util.types import AttributeValue
 from pydantic import BaseModel
 
 from lilypad.models import FnParamsPublic, VersionPublic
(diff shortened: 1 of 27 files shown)


lilypad/_utils.py
```diff
@@ -103,37 +100,41 @@ def get_llm_function_version(
...
-    try:
-        llm_fn_version = lilypad_client.get_llm_function_by_hash(hash)
-    except client.NotFoundError:
-        llm_fn_version = lilypad_client.post_llm_function(
-            function_name=fn.__name__,
-            code=code,
-            version_hash=hash,
-            arg_types=arg_types,
-        )
+    if not synced:
```
willbakst, line 103: is this going to affect switching from non-synced to synced for the same function name?

lilypad/llm_fn.py
```diff
@@ -70,26 +85,65 @@ def decorator(
...
+                        version=version.version,
+                    )
                     return cast(_R, await trace_decorator(fn)(*args, **kwargs))
 
                 if not version.fn_params:
                     raise ValueError(f"Synced function {fn.__name__} has no params.")
 
+                fn_params = version.fn_params
+
+                @contextmanager
```
willbakst, line 108: this looks exactly the same as the one below and doesn't seem to rely on sync vs. async, can we convert this into a utility that returns the context manager instead?

lilypad/llm_fn.py
```diff
@@ -100,26 +154,76 @@ async def inner_async(*args: _P.args, **kwargs: _P.kwargs) -> _R:
...
+                            if version.fn_params
+                            else "",
+                            "lilypad.is_async": False,
+                        }
+                        span.set_attributes(attributes)
+                        yield span
 
                 if not synced:
-                    return cast(_R, trace_decorator(fn)(*args, **kwargs))
+                    if hasattr(fn, "__mirascope_call__"):
```
willbakst, line 185: why do I only see this check for sync and not async?

lilypad/llm_fn.py
```diff
@@ -100,26 +154,76 @@ async def inner_async(*args: _P.args, **kwargs: _P.kwargs) -> _R:
...
+                        decorator = middleware_factory(
```
willbakst, line 186: i think we could structure this better so we only use middleware_factory once.

For example:

```python
is_mirascope_call = hasattr(fn, "__mirascope_call__")
if not synced and not is_mirascope_call:
    decorator = trace(...)
    return cast(_R, decorator(fn)(*args, **kwargs))
decorator = middleware_factory(...)
if not synced and is_mirascope_call:
    return cast(_R, decorator(fn)(*args, **kwargs))

if not version.fn_params:
    ...

return traced_synced_llm_function_constructor(...)...
```

WDYT?

pyproject.toml
```diff
@@ -31,7 +31,7 @@ classifiers = [
     "Topic :: Software Development :: Libraries",
 ]
 dependencies = [
-    "mirascope>=1.2.0",
+    "mirascope>=1.5.2",
```
willbakst, line 34: nit: i believe this will now need `>=1.6.0` because of another bug fix in that release for using built-in response models

lilypad/_utils.py
brenkao: the traces if you switch from non-synced to synced will be different under the same function name

lilypad/llm_fn.py
brenkao: done

lilypad/llm_fn.py
brenkao: fixed

pyproject.toml
brenkao: changed to 1.6.1

PR #22, opened 2024-11-13
feat: add unitest for prompts.py

koxudaxi:


lilypad/_utils/closure.py
```diff
@@ -184,7 +185,9 @@ def _collect_function_and_deps(self, func: Callable[..., Any]) -> None:
                 return
 
             # Get function source and create AST
-            source = inspect.getsource(func)
+            raw_source = inspect.getsource(func)
+            source = textwrap.dedent(raw_source)
```
koxudaxi, line 189: clean up indent to avoid syntax error.

tests/test_prompts.py
```diff
@@ -0,0 +1,155 @@
...
+    response = recommend_book("mystery")
+    assert isinstance(response, str)
+    assert response == "Mocked book recommendation"
+    assert len(tracking) == 1
+    assert tracking[0]["genre"] == "mystery"
+    mock_prompt_client.get_prompt_active_version.assert_called_once()
+    mock_create_mirascope_call.assert_called()
+
+
+@pytest.mark.slow
```
willbakst, line 140: can you explain the purpose of this test?

willbakst (approved):

tests/test_prompts.py
koxudaxi: When I first wrote the test, mock was incomplete and slow, so I added this mark so it could be skipped. However, since mock worked fine, this flag is not necessary.

PR #24, opened 2024-11-14
feat: add unitest for models

koxudaxi:


lilypad/server/models/spans.py
```diff
@@ -71,7 +71,8 @@ def _convert_span_table_to_public(
         elif span.scope == Scope.LLM:
             data = span.data
             display_name = f"{data['attributes']['gen_ai.system']} with '{data['attributes']['gen_ai.request.model']}'"
-        else:
+        else: # pragma: no cover
+            # This should never happen. Because the scope is an Enum, it should always be one of the two values. And, SQLModel should enforce this.
```
willbakst, line 75: if this is the case, why don't we just put the above `elif` block inside an `else` block instead?

tests/server/test_models.py
```diff
@@ -0,0 +1,296 @@
...
+    SpanPublic,
+    SpanTable,
+    VersionCreate,
+    VersionPublic,
+    VersionTable,
+)
+
+
+@pytest.fixture
+def engine():
```
willbakst, line 32: nit: rename to `fixture_engine` so it's clear it's a fixture

tests/server/test_models.py
```diff
@@ -0,0 +1,296 @@
...
+    """Create an in-memory database engine for testing."""
+    engine = create_engine(
+        "sqlite://", connect_args={"check_same_thread": False}, poolclass=StaticPool
+    )
+    SQLModel.metadata.create_all(engine)
+    return engine
+
+
+@pytest.fixture
+def session(engine):
```
willbakst, line 42: nit: same here, `fixture_session`

tests/server/test_models.py
```diff
@@ -0,0 +1,296 @@
...
+    """Create an in-memory database engine for testing."""
+    engine = create_engine(
+        "sqlite://", connect_args={"check_same_thread": False}, poolclass=StaticPool
+    )
+    SQLModel.metadata.create_all(engine)
+    return engine
+
+
+@pytest.fixture
+def session(fixture_engine):
```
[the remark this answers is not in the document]
willbakst: Circling back on this, I think this should actually just be `session` and the above should just be `engine`

For fixtures that use mocks, we should prepend mock e.g. `mock_session`, but this particular engine and session are not actually mocks they are real, so using mock would be confusing.

tests/server/test_models.py
```diff
@@ -0,0 +1,296 @@
...
+    SpanPublic,
+    SpanTable,
+    VersionCreate,
+    VersionPublic,
+    VersionTable,
+)
+
+
+@pytest.fixture
+def fixture_engine():
```
willbakst, line 32: let's actually change this back to just `engine`

willbakst: The convention should be:
1. Use `mock_` when the fixture actually mocks something (e.g. fake data, MagicMock, etc.)
2. Use clear naming for anything that is not a mock (e.g. `sqlite_engine` in this case)

willbakst, line 32: this is missing a type hint return annotation

tests/server/test_models.py
willbakst, line 42: this is missing a type hint on the engine arg

tests/server/test_models.py
```diff
@@ -0,0 +1,296 @@
...
+
+@pytest.fixture
+def session(fixture_engine):
+    """Create a new database session for testing."""
+    with Session(fixture_engine) as session:
+        yield session
+        session.rollback()
+
+
+def test_base_sql_model():
```
willbakst, line 49: these are all missing return annotations of `None` please update all tests to return `None`

willbakst (approved):

PR #25, opened 2024-11-14
feat: Add unittest for _utils

koxudaxi:


tests/_utils/test_closure.py
```diff
@@ -228,3 +239,361 @@ def _recommend_book_series_from_available(genre: str) -> openai.OpenAIDynamicCon
...
+        obj = SomeClass()
+        obj.method()
+
+    closure, _ = compute_closure(func)
+    assert "SomeClass" in closure
+    assert "def func()" in closure
+
+
+def test_get_code_location_no_module():
+    """Test that _get_code_location handles functions with no module."""
```
willbakst, line 487: this says "handles" but it appears it actually just throws a validation error?

tests/_utils/test_closure.py
```diff
@@ -228,3 +239,361 @@ def _recommend_book_series_from_available(genre: str) -> openai.OpenAIDynamicCon
...
+    assert not collector._is_serializable_value(object())
+
+
+def test_collect_with_type_aliases():
+    """Test handling of type aliases in annotations."""
+    from typing import TypeAlias
+
+    MyType: TypeAlias = str  # pyright: ignore [reportGeneralTypeIssues]
+
+    def func(param: MyType) -> MyType:
```
willbakst, line 533: hmmmm, this is interesting

if the user were to switch the MyType alias from `str` to `int` then the closure and hash would remain the same even though they should change

the same is also true if the user updates from `MyType` to `str` which should have the same closure and hash but wont

I'm thinking for this case we should actually resolve the alias and use the original definition (i.e. `str` here) for the closure rather than the alias so that the closure/hash versioning is more consistent and correct

WDYT?

tests/_utils/test_functions.py
```diff
@@ -0,0 +1,1135 @@
...
+from openai.types.chat.chat_completion import Choice
+from pydantic import BaseModel, computed_field
+from pydantic.json_schema import SkipJsonSchema
+
+from lilypad import Message
+from lilypad._utils.functions import create_mirascope_call, inspect_arguments
+from lilypad.server.models import Provider
+
+
+class MockResponse(
```
willbakst, line 29: have we already defined this for messages.py? if so, I think maybe we should make this a fixture and put it in a conftest?

tests/_utils/test_functions.py
```diff
@@ -0,0 +1,1135 @@
...
+            )
+        ],
+        created=0,
+        model="gpt-4",
+        object="chat.completion",
+        system_fingerprint="test-fingerprint",
+    )
+
+
+class MockBaseTool(BaseTool[Any], ABC):
```
willbakst, line 168: same thing here and for all of these types (which i'm fairly certain we already defined), I think for instances of mocks like this we should pull them out into a conftest if they are used in multiple test locations

tests/_utils/test_closure.py
```diff
@@ -228,3 +239,361 @@ def _recommend_book_series_from_available(genre: str) -> openai.OpenAIDynamicCon
...
+    module = inspect.getmodule(test_func)
+    with patch.object(module, "__file__", None):
+        location = collector._get_code_location(test_func)
+        assert location.module_path == ""
+
+
+def test_is_user_defined_import():
+    """Test the _is_user_defined_import method."""
+    collector = _DependencyCollector()
+    assert collector._is_user_defined_import("from typing import List") is True
```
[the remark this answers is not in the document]
koxudaxi: ummm, I thought this is special case.
https://github.com/Mirascope/lilypad/blob/c6a58aa079b1fe7103f3e926d471543dc5bbb8e2/lilypad/_utils/closure.py#L307-L310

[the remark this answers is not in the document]
willbakst: oh hmm, that seems wrong lol

we should remove that, right? that's definitely not a user defined import...

PR #26, opened 2024-11-20
feat: initial github oauth authentication

brenkao:


lilypad/cli/commands/auth.py
```diff
@@ -0,0 +1,87 @@
...
+
+import httpx
+import typer
+
+from ...server.models import DeviceCodeTable
+
+AUTH_URL = "https://your-auth-endpoint.com/auth"
+REDIRECT_PORT = 5173
+LOGIN_URL = "http://localhost:5173/auth/login"
+API_BASE_URL = "http://localhost:8000/api/v0"  # Your FastAPI server
```
willbakst, line 18: nit: remove comment?

lilypad/cli/commands/auth.py
```diff
@@ -0,0 +1,87 @@
...
+import webbrowser
+
+import httpx
+import typer
+
+from ...server.models import DeviceCodeTable
+
+AUTH_URL = "https://your-auth-endpoint.com/auth"
+REDIRECT_PORT = 5173
+LOGIN_URL = "http://localhost:5173/auth/login"
```
willbakst, line 17: is this the right route? wouldn't this be the same as the api in production?

lilypad/cli/commands/auth.py
```diff
@@ -0,0 +1,87 @@
...
+API_BASE_URL = "http://localhost:8000/api/v0"  # Your FastAPI server
+
+
+def generate_device_code() -> str:
```
willbakst, line 21: nit: we should make this and the other functions private to this command

lilypad/cli/commands/local.py
```diff
@@ -125,15 +126,18 @@ def signal_handler(sig: int, frame: FrameType | None) -> None:
...
                 os.mkdir(".lilypad")
+                project = lilypad_client.post_project(project_name)
                 with open(".lilypad/config.json", "w") as f:
-                    data = {
-                        "project_name": project_name,
-                        "project_id": project.id,
-                        "port": new_port,
-                    }
-                    json.dump(data, f, indent=4)
+                    json.dump(
```
willbakst, line 132: nit: i like the previous formatting with data set outside of the json dump

lilypad/server/services/device_codes.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import Annotated
+
+from fastapi import Depends, HTTPException, status
+from sqlmodel import Session, select
+
+from ..db import get_session
+from ..models import DeviceCodeTable
+
+
+class DeviceCodeService:
```
willbakst, line 12: is there a reason this doesn't use the base service?

lilypad/server/models/auth.py
```diff
@@ -0,0 +1,38 @@
...
+class UserBase(BaseModel):
+    """Base user model"""
+
+    first_name: str | None = None
+    raw_profile: dict
+    session_id: str
+    expires_at: str
+    access_token: str | None = None
+    organization_id: str | None = None
+    organizations: Sequence[Organization] | None = None
```
willbakst, line 26: could we organize all the WorkOS stuff in it's own folders (e.g. `/auth/workos` or something) and then construct the models and API and such based on an environment variable (e.g. `AUTH_PROVIDER`)?

The flip side to this question is that if we can make this change later without it being a breaking change then we can just ignore for now.

brenkao: it wont ever be a breaking change, since the user will never see this stuff. for example, in the client we would toggle. between workos and other providers, etc and we would change the url internally but that wont change the flow. get authorization url -> user authorizes -> redirect to client -> redirect to server -> sign token -> send token to user -> user uses token for requests.

brenkao: although putting this in a workos folder seems to make sense so i will do that anyways

lilypad/cli/commands/auth.py
brenkao: 1. this is the route for local
2. this need to change to be env that user sets since it will need to point to users domain

willbakst: but when they run `lilypad local` it will serve everything from port 8000 (vs. port 5173 which is the vite dev port)?

lilypad/server/services/device_codes.py
willbakst: ok, that makes sense.

we should probably make a new `UnauthenticatedBaseService` or something in the future should the need arise for other services that can't have it

lilypad/server/models/auth.py
willbakst: yeah, that makes sense. I think really I just want to do any easy work now to prep for future additional providers, but doesn't matter too much in the short term if we can introduce those changes without it breaking.

My main question is if we should have something like separate user tables partitioned by provider and then indicate which user table to use based on the environment variable or if we should have a single user table that we instantiate based on the provider. I feel like the former is more future proof?

lilypad/cli/commands/auth.py
brenkao: makes sense good eye

lilypad/server/services/base.py
```diff
@@ -24,6 +25,7 @@ def find_record_by_id(self, id: int | str) -> _TableT:
         """Find record by id"""
         record_table = self.session.exec(
             select(self.table).where(
+                self.table.organization_uuid == self.user.active_organization_uuid,  # pyright: ignore[reportAttributeAccessIssue]
```
willbakst, line 28: is there an issue here because `self.user` is not defined? can we define it so there isn't a type error?

I see it in the init, why not include it as an attribute of the class?

lilypad/server/client.py
```diff
@@ -58,9 +60,29 @@ def __init__(
         if headers:
             self.session.headers.update(headers)
 
+        self._token = None
+        if token:
+            self.token = token
```
willbakst, line 65: this seems redundant? why not just `self.token = token if token else None`?

lilypad/_configure.py
```diff
@@ -23,7 +23,13 @@ class _JSONSpanExporter(SpanExporter):
 
     def __init__(self, base_url: str) -> None:
         """Initialize the exporter with the custom endpoint URL."""
-        self.client = LilypadClient(base_url, timeout=10)
+        config = load_config()
+
+        self.client = LilypadClient(
+            base_url,
```
willbakst, line 29: should we load `base_url` from the config / environment?

client/src/auth.tsx
```diff
@@ -0,0 +1,78 @@
...
+const saveToStorage = (session: UserPublic | null) => {
+  if (session) {
+    localStorage.setItem(AUTH_STORAGE_KEY, JSON.stringify(session));
+  } else {
+    localStorage.removeItem(AUTH_STORAGE_KEY);
+  }
+};
+
+const loadFromStorage = (): UserPublic | null => {
+  // if (import.meta.env.DEV)
```
willbakst, line 28: should this remain commented for later or should we do something about this before merging?

client/src/auth.tsx
```diff
@@ -0,0 +1,78 @@
...
+  //     first_name: "Local User",
+  //     raw_profile: {},
+  //     expires_at: "",
+  //   };
+  const stored = localStorage.getItem(AUTH_STORAGE_KEY);
+  if (!stored) return null;
+
+  try {
+    const session = JSON.parse(stored) as UserPublic;
+    // TODO: Check if token is expired
```
willbakst, line 39: same here? leaving as todo or should we resolve?

client/src/components/GithubLogin.tsx
```diff
@@ -0,0 +1,56 @@
...
+      {children}
+    </Button>
+  );
+};
+
+export const GithubLogin = () => {
+  const getAuthUrl = () => {
+    const githubAuthUrl = "https://github.com/login/oauth/authorize";
+    const params = new URLSearchParams({
+      client_id: import.meta.env.VITE_GITHUB_CLIENT_ID,
```
willbakst, line 39: are we going to be able to read these from the server on build?

lilypad/cli/commands/auth.py
```diff
@@ -0,0 +1,147 @@
...
+        try:
+            response = await client.delete(
+                f"{settings.api_url}/v0/device-codes/{device_code}"
+            )
+            return response.status_code == 200
+        except httpx.RequestError:
+            return False
+
+
+def show_project_selection(projects: list[ProjectPublic]) -> ProjectPublic:
```
willbakst, line 69: nit: make private?

lilypad/cli/commands/auth.py
```diff
@@ -0,0 +1,147 @@
...
+    """Show a list of projects and prompt the user to select one."""
+    print("\nAvailable projects:")
+    for idx, project in enumerate(projects, 1):
+        print(f"[green]{idx}[/green]. {project.name}")
+
+    choice = IntPrompt.ask("Select project number")
+    return projects[choice - 1]
+
+
+def check_existing_token(settings: Settings) -> bool:
```
willbakst, line 79: nit: make private?

lilypad/server/_utils/auth.py
```diff
@@ -0,0 +1,105 @@
...
+    return jwt.encode(
+        json.loads(user_data.model_dump_json()),
+        settings.jwt_secret,
+        algorithm=settings.jwt_algorithm,
+    )
+
+
+if os.getenv("LILYPAD_LOCAL", "false").lower() == "true":
+
+    async def oauth2_scheme(token: str | None = None) -> str:  # pyright: ignore[reportRedeclaration]
```
willbakst, line 39: nit: can we avoid the type ignore if we name this something like `get_local_token` and set `oauth2_scheme = get_local_token`?

lilypad/server/api/v0/auth/github_api.py
```diff
@@ -0,0 +1,115 @@
...
+    UserRole,
+    UserTable,
+)
+from ....services import DeviceCodeService
+from ....settings import Settings, get_settings
+
+github_router = APIRouter()
+
+
+@github_router.get("/github/callback", response_model=UserPublic)
```
willbakst, line 25: question: this is just for `lilypad auth` right? we don't use this on the client?

lilypad/server/models/user_organizations.py
willbakst, line 1: just to check my understanding, this is for permissions, right?

I'm wondering if there is a better naming convention to make this more clear? Would something like `UserPermissions` be better?

lilypad/server/services/base.py
brenkao: type error is on self.table where organization_uuid is not defined. every table would need to extend from a base model with organization_uuid

lilypad/server/_utils/auth.py
brenkao: i tried what you did and still got an error, so not sure what the best way to handle this is.

lilypad/server/api/v0/auth/github_api.py
brenkao: its used for both. if user does lilypad auth for the first time without an account, it would hit the callback endpoint. the 2nd use case is if the user is already logged in, we redirect directly to the projects page, so lilypad auth needs to handle it there too

lilypad/server/models/user_organizations.py
brenkao: this is standard table naming conventions in sql. the reason its not userpermissions is because if we add more columns, then its no longer userpermissions

client/src/components/GithubLogin.tsx
brenkao: yes, here are the values if you are curious
```
DEV: false
MODE: "production"
PROD: true
```

lilypad/server/services/base.py
willbakst: should we not do that? seems like here we're expecting every model to have it? this will crash if not, right?

lilypad/server/_utils/auth.py
willbakst: ok just leave it as-is for now

lilypad/server/models/user_organizations.py
willbakst: got it

lilypad/server/services/base.py
willbakst: online discussion: leaving as-is and fixing in the next PR

willbakst (approved):

PR #29, opened 2024-11-25
feat: add unittest for server

koxudaxi:


lilypad/server/services/versions.py
```diff
@@ -88,19 +87,34 @@ def find_prompt_active_version(
...
+        current_active_versions = self.session.exec(stmt).all()
+
+        for version in current_active_versions:
+            version.is_active = False
+            self.session.add(version)
+
+        # Activate the new version
         new_active_version.is_active = True
         self.session.add(new_active_version)
-        self.session.flush()
```
willbakst, line 102: @brenkao i remember using `flush` for a reason here but it's eluding me

brenkao: flush is if you want to grab the updated id, we already commit everything at the end so committing is redundant

brenkao: or if there is an exception in the api, then we don't commit the transaction. if we call commit here, then we will always save, even if there are errors elsewhere

willbakst: ah ok yes this is what I was thinking of thank you

@koxudaxi please take this into account with this update so that we properly handle potential errors with flush instead of commit

brenkao: we should be able to keep flush, but add refresh and that should be good

koxudaxi: Thank you for the commit.
I reverted `commit` to `flush()`

willbakst (approved):

PR #33, opened 2024-12-03
feat: add env for external db, fix: postgres compatibility, various bugs with uuid.

brenkao:


lilypad/cli/commands/local.py
```diff
@@ -33,8 +32,19 @@ def _start_lilypad(project_dir: Path, port: str | int) -> subprocess.Popen:
...
     # Sets the LILYPAD_PROJECT_DIR so Lilypad knows where to look for the database
     env["LILYPAD_PROJECT_DIR"] = str(project_dir)
-    env["LILYPAD_LOCAL"] = "true"
+    env["LILYPAD_ENVIRONMENT"] = "local"
     files_dir = files("lilypad").joinpath("server")
+    if not os.path.exists("pad.db"):
+        subprocess.run(
+            [
+                "python",
+                "db/setup.py",
```
willbakst, line 41: will this run properly in a user's environment when they install lilypad? previously we just used the `create_tables` method, was there something wrong with that?

lilypad/server/api/v0/versions_api.py
```diff
@@ -40,7 +41,7 @@ async def get_version_by_uuid(
 
 
 @versions_router.get(
-    "/projects/{project_uuid}/functions/{function_name}/versions",
+    "/projects/{project_uuid}/functions/name/{function_name}/versions",
```
willbakst, line 44: what sparked the inclusion of `/name/`?

lilypad/server/client.py
```diff
@@ -59,11 +59,11 @@ def __init__(
             )
         except FileNotFoundError:
             self.project_uuid = None
-
         if headers:
             self.session.headers.update(headers)
 
-        self._token = token if token else None
```
willbakst, line 66: what's going on here?

lilypad/server/db/session.py
```diff
@@ -5,10 +5,17 @@
...
+from ..settings import get_settings
+
+settings = get_settings()
+if settings.environment == "local" or settings.environment == "test":
+    db_directory = os.getenv("LILYPAD_PROJECT_DIR", os.getcwd())
+    engine = create_engine(
+        f"sqlite:///{db_directory}/pad.db", connect_args={"check_same_thread": False}
+    )
+else:
+    database_url = f"postgresql://{settings.db_user}:{settings.db_password}@{settings.db_host}:{settings.db_port}/{settings.db_name}"
```
willbakst, line 17: i'm assuming this means for now we are only going to accept postgres connections for non-local stuff?

lilypad/server/settings.py
```diff
@@ -18,19 +18,25 @@ class Settings(BaseSettings):
...
     def config(self) -> dict[str, Any]:
         """Get the configuration for the current environment"""
         configs = {
             "development": {
                 "api_url": "http://localhost:8000/api",
-                "client_url": "http://localhost:5173",
+                "client_url": "",
+            },
+            "local": {
+                "api_url": "http://localhost:8000/api",
```
willbakst, line 38: shouldn't the port be configurable?

lilypad/cli/commands/local.py
brenkao: i can test in a fresh environment. the issue with the previous method was that i had no way of knowing which environment they were in. so with the new method, we set the environment to local, instead of it using the default which is dev.

lilypad/server/api/v0/versions_api.py
brenkao: name conflicts with the other route that got changed. the issue is that we had two routes with the same path but one was using id (int) and the other with (str). that was bad practice which caused a bug when we changed to UUID, which gets converted to str and caused issues

lilypad/server/client.py
brenkao: when i set lilypad_client on initialization with token, it wasnt properly setting the authorization bearer token, so i had to update this

lilypad/server/db/session.py
brenkao: yes, unless you want another env for dialect

lilypad/server/settings.py
brenkao: hmm yeah, i will add a new env that is backend port

lilypad/cli/commands/local.py
willbakst: could we update `create_tables` to take the environment as an argument?

lilypad/server/api/v0/versions_api.py
willbakst: got it

lilypad/server/db/session.py
willbakst: i think this is fine for now we can always add more later but since we're starting with just postgres we can hold off

lilypad/server/client.py
willbakst: i still don't quite understand why we need both `self._token` and `self.token` is `self._token` will always be `None`?

lilypad/_utils/middleware.py
```diff
@@ -38,10 +38,8 @@ def custom_context_manager(
         fn: Callable,
     ) -> Generator[Span, Any, None]:
         tracer = get_tracer("lilypad")
-        config = load_config()
-        lilypad_client = LilypadClient(
-            base_url=f"http://localhost:{config.get('port', 8000)}/api", timeout=10
-        )
+        settings = get_settings()
+        lilypad_client = LilypadClient(base_url=settings.api_url, timeout=10)
```
willbakst, line 42: nit: why are we calling it `api_url` instead of `base_url`?

lilypad/prompts.py
```diff
@@ -12,14 +12,16 @@
...
 from .server.client import LilypadClient
+from .server.settings import get_settings
 
 _P = ParamSpec("_P")
 _R = TypeVar("_R")
 
 config = load_config()
+settings = get_settings()
 
 lilypad_client = LilypadClient(
```
willbakst, line 23: nit: seems to me like we should add `api` to the `base_url` inside of `LilypadClient` and have `settings.base_url`?

lilypad/server/api/v0/auth/github_api.py
```diff
@@ -48,7 +48,7 @@ async def github_callback(
                     "client_id": settings.github_client_id,
                     "client_secret": settings.github_client_secret,
                     "code": code,
-                    "redirect_uri": settings.github_redirect_uri,
+                    "redirect_uri": f"{settings.client_url}/auth/callback",
```
willbakst, line 51: aha, is `client_url` different from `base_url`?

lilypad/server/settings.py
```diff
@@ -35,12 +42,12 @@ def config(self) -> dict[str, Any]:
                 "client_url": "",
             },
             "local": {
-                "api_url": "http://localhost:8000/api",
-                "client_url": "http://localhost:8000",
+                "api_url": self.local_api_url,
```
willbakst, line 45: seems weird here to have `api_url` and `client_url` when it's really just `base_url` and then the inclusion of `/api`?

also curious where we're handling possible updates to the api version that's currently `v0` or if we just handle that in the client class?

client/src/components/GithubLogin.tsx
```diff
@@ -38,7 +38,7 @@ export const GithubLogin = ({ deviceCode }: { deviceCode?: string }) => {
 
     const params = new URLSearchParams({
       client_id: import.meta.env.VITE_GITHUB_CLIENT_ID,
-      redirect_uri: import.meta.env.VITE_GITHUB_REDIRECT_URI,
+      redirect_uri: `${import.meta.env.VITE_API_SERVER}/auth/callback`,
```
willbakst, line 41: wait so can you clarify where the VITE environment is getting set? I thought we agreed that the environment for the static client should be loaded from the server?

lilypad/server/api/v0/auth/github_api.py
brenkao: yes, it can be.

lilypad/server/settings.py
brenkao: we will handle that in the client when we get to v1

client/src/components/GithubLogin.tsx
brenkao: as discussed, this will be done at a later pr. for now, it is getting set in your .zshrc env

lilypad/prompts.py
brenkao: updated to base_url

willbakst (approved):

PR #34, opened 2024-12-05
`uv` run function closure

willbakst:
> [!WARNING]
> The `Closure.run` method should not be used without careful consideration since it will run the code provided. Make sure that if you're running a closure that the code is safe. Going forward, we will need to ensure we only ever run the code of a closure in a safe sandbox. As it's currently implemented, the `run` method should only be used for development and testing until we figure out the sandboxing.


lilypad/closure.py
```diff
@@ -0,0 +1,543 @@
...
+    extras: list[str] | None
+
+
+def _is_third_party(module: ModuleType, site_packages: set[str]) -> bool:
+    module_file = getattr(module, "__file__", None)
+    return (
+        module.__name__ in sys.stdlib_module_names
+        or module_file is None
+        or any(
+            str(Path(module_file).resolve()).startswith(site_pkg)
```
koxudaxi, line 37: I don’t have access to a Windows environment to verify, but it’s likely that this would fail to match due to differences in the case of drive letters.
In the future, should we consider adding Windows or macOS to the CI pipeline? Since this involves path manipulation, it might be necessary.

lilypad/closure.py
```diff
@@ -0,0 +1,543 @@
...
+            check=True,
+            capture_output=True,
+            text=True,
+        )
+        subprocess.run(
+            ["ruff", "format", "--line-length=88", str(tmp_path)],
+            check=True,
+            capture_output=True,
+            text=True,
+        )
```
koxudaxi, line 443: Although I can’t think of a concrete scenario, there is a possibility that the generated code could change if ruff detects configuration files like ruff.toml or pyproject.toml.
Should we specify certain options to ensure the results remain consistent?
If this concern is not relevant, feel free to ignore this comment.

lilypad/closure.py
```diff
@@ -0,0 +1,543 @@
...
+                    alias = parts[parts.index("as") + 1]
+                    self.alias_mapping[alias] = original_name
+
+    def visit_Attribute(self, node: ast.Attribute) -> ast.AST:
+        names = []
+        current = node
+        while isinstance(current, ast.Attribute):
+            names.append(current.attr)
+            current = current.value
+        if isinstance(current, ast.Name) and current.id != "self":
```
koxudaxi, line 237: check only `self`? Should we ignore `cls` too?

willbakst: good catch! fixed

lilypad/closure.py
willbakst: I've added Windows to our actions to see if this still works. I'm opting to skip macOS because we run those tests on macOS locally anyway (and macOS minutes are 10x so we'll hit limits faster and likely have to pay)

brenkao (approved):

PR #35, opened 2024-12-06
Hosted lilypad

brenkao:


lilypad/cli/commands/__init__.py
```diff
@@ -1,6 +1,7 @@
 """Commands for the `lilypad` CLI."""
 
+from ._utils import get_and_create_config
 from .auth import auth_command
 from .local import local_command
 
-__all__ = ["auth_command", "local_command"]
+__all__ = ["auth_command", "get_and_create_config", "local_command"]
```
willbakst, line 7: why is this private utility exported as a command?

brenkao: oops, it used to not be private, and forgot to delete

willbakst (approved):

PR #38, opened 2024-12-11
Generation updates

willbakst:


lilypad/_utils/middleware.py
```diff
@@ -44,17 +44,14 @@ def custom_context_manager(
...
-                if version.prompt
-                else prompt_template or "",
+                "lilypad.type": "generation",
+                "lilypad.generation.uuid": str(version.uuid),
+                "lilypad.generation.name": fn.__name__,
+                "lilypad.generation.signature": version.signature,
+                "lilypad.generation.code": version.code,
+                "lilypad.generation.arg_types": json.dumps(arg_types),
+                "lilypad.generation.arg_values": json.dumps(arg_values),
+                "lilypad.generation.prompt_template": prompt_template or "",
```
brenkao, line 54: why does generation have prompt template, shouldnt this be lilypad.prompt.template?

lilypad/generations.py
```diff
@@ -71,23 +94,13 @@ async def inner_async(*args: _P.args, **kwargs: _P.kwargs) -> _R:
                     f"{fn.__name__}"
                 ) as span:
                     output = await fn(*args, **kwargs)
-                    results = str(output)
                     if isinstance(output, BaseModel):
```
brenkao, line 97: move this section inside `_contruct_trace_attributes` since its shared for async

lilypad/prompts.py
```diff
@@ -25,31 +28,234 @@
...
+                convert_common_call_params,
+            )
+
+            return convert_common_call_params(self.common_call_params)
+        else:
+            raise NotImplementedError(f"Unknown provider: {provider}")
+
+
+def _construct_trace_attributes(
+    version: PromptPublic,
```
brenkao, line 156: rename prompt for variable?

brenkao (approved):

lilypad/_utils/middleware.py
willbakst: this is for mirascope calls that will have prompt templates at the generation level for our first class support

lilypad/prompts.py
willbakst: i used `version` everywhere (for both prompt and generation) since it's the version we're using. should i instead use `prompt` / `generation` or `prompt_version` / `generation_version`?

brenkao: i like prompt_version / generation_version a little better than version everywhere, even though its in the prompts file

lilypad/_utils/middleware.py
brenkao: got it

lilypad/server/models/generations.py
```diff
@@ -0,0 +1,62 @@
...
+    hash: str = Field(nullable=False, index=True, unique=True)
+    dependencies: dict[str, DependencyInfo] = Field(
+        sa_column=Column(JSON), default_factory=dict
+    )
+
+
+class GenerationCreate(Closure):
+    """Generation create model."""
+
+    arg_types: dict[str, str]
```
brenkao, line 44: `arg_types` is only on create, but not on the GenerationTable so this column will never be created. is this intended?

lilypad/server/models/generations.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class GenerationPublic(_GenerationBase):
+    """Generation public model."""
+
+    uuid: UUID
```
brenkao, line 50: missing prompt on public model

lilypad/server/models/generations.py
willbakst: done

lilypad/server/models/generations.py
willbakst: done

lilypad/server/models/generations.py
```diff
@@ -0,0 +1,62 @@
...
+    name: str = Field(nullable=False, index=True, min_length=1)
+    signature: str = Field(nullable=False)
+    code: str = Field(nullable=False)
+    hash: str = Field(nullable=False, index=True, unique=True)
+    dependencies: dict[str, DependencyInfo] = Field(
+        sa_column=Column(JSON), default_factory=dict
+    )
+
+
+class GenerationCreate(Closure):
```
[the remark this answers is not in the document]
willbakst: I've removed this dependency and instead made it depend on the base

PR #49, opened 2024-12-16
feat: UI updates generations

brenkao:


lilypad/server/services/prompts.py
```diff
@@ -15,13 +16,63 @@ class PromptService(BaseService[PromptTable, PromptCreate]):
     table: type[PromptTable] = PromptTable
     create_model: type[PromptCreate] = PromptCreate
 
+    def find_records_by_name(
```
willbakst, line 19: nit: should this be `find_prompts_by_name`?

brenkao: yeah i also updated another one like this

willbakst (approved):

PR #54, opened 2024-12-17
fix: various bugs with editor and version

brenkao:


client/src/components/GenerationSpans.tsx
```diff
@@ -16,5 +16,6 @@ export const GenerationSpans = ({
   const { data } = useSuspenseQuery(
     spansByGenerationQueryOptions(projectUuid, generationUuid)
   );
+  console.log(data);
```
willbakst, line 19: remove?

lilypad/server/services/prompts.py
```diff
@@ -29,6 +29,18 @@ def find_prompts_by_name(
         ).all()
         return record_tables
 
+    def get_next_version(self, project_uuid: UUID, name: str) -> int:
```
willbakst, line 32: nit: should we put this on the BaseOrganizationSQLModel?

lilypad/server/api/v0/prompts_api.py
```diff
@@ -203,6 +207,7 @@ def run_version(
 
 
 def _run_playground(code: str) -> str:
+    print(code)
```
willbakst, line 210: nit: remove

lilypad/generations.py
```diff
@@ -73,6 +73,9 @@ def _construct_trace_attributes(
         "lilypad.generation.arg_values": json.dumps(arg_values),
         "lilypad.generation.prompt_template": prompt_template,
         "lilypad.generation.output": str(output),
+        "lilypad.generation.version": generation.version_num
+        if generation.version_num
+        else 0,
```
willbakst, line 78: does `0` mean no version number? why not `-1`?

lilypad/server/services/prompts.py
brenkao: no, there are many models without version

lilypad/generations.py
brenkao: mainly for falsey but i can do -1

willbakst (approved):

Issue #66, opened 2024-12-18
Create new projects from the UI


willbakst:
### Description

We should be able to create a new project in the UI and let the user copy the UUID. This would make it possible to bypass the `lilypad auth` command simply by setting environment variables.

PR #90, opened 2025-01-07
feat: added api key authentication

### LINKED ISSUE #70: Update authentication flow to use an API key
### Description

We should enable creating API keys as part of a project. This way users can simply set e.g. `LILYPAD_API_KEY` in their environment and have it point everything to the correct project.

We may still want to require Project name (i.e. `LILYPAD_PROJECT`) just to be safe (e.g. prevent accidentally sending generations to the wrong project).

This should make the `lilypad auth` command obsolete

brenkao:
Closes #70 #66


lilypad/cli/commands/auth.py
```diff
@@ -126,6 +127,10 @@ def auth_command(
 ) -> None:
     """Open browser for authentication and save the received token."""
     settings = get_settings()
+    warnings.warn(
```
willbakst, line 130: why not run this first before anything else?

lilypad/server/api/v0/api_keys_api.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+@api_keys_api.delete("/api-keys/{api_key_uuid}")
+async def delete_api_key(
+    api_key_uuid: UUID,
+    api_key_service: Annotated[APIKeyService, Depends(APIKeyService)],
+) -> bool:
+    """Delete an API key."""
+    try:
+        api_key_service.delete_record_by_uuid(api_key_uuid)
```
willbakst, line 44: I feel like the exception should be cause inside of the base deletion service? then we can just return `delete_record_by_uuid` which we have return the success/failure bool

lilypad/server/migrations/versions/0002_added_api_key_table.py
willbakst, line 1: nit: this migration may need resolution with another from the other pr I just reviewed from you

lilypad/server/models/api_keys.py
```diff
@@ -0,0 +1,61 @@
...
+    project_uuid: UUID = Field(index=True, foreign_key=f"{PROJECT_TABLE_NAME}.uuid")
+
+
+class APIKeyTable(_APIKeyBase, BaseOrganizationSQLModel, table=True):
+    """APIKey table."""
+
+    __tablename__ = API_KEY_TABLE_NAME  # type: ignore
+    key_hash: str = Field(nullable=False)
+    user_uuid: UUID = Field(index=True, foreign_key=f"{USER_TABLE_NAME}.uuid")
+    organization: "OrganizationTable" = Relationship(back_populates="api_keys")
```
willbakst, line 38: why don't we have an `organization_uuid` on the `_APIKeyBase` for this relationship? we have it for `user` and `project`? Is it because we don't surface `OgranizationPublic` on `APIKeyPublic`?

lilypad/server/models/projects.py
```diff
@@ -37,6 +39,7 @@ class ProjectPublic(_ProjectBase):
     generations: list[GenerationPublic] = []
     prompts: list[PromptPublic] = []
     response_models: list[ResponseModelPublic] = []
+    created_at: datetime
```
willbakst, line 42: i'm assuming we should also then surface `updated_at` so that we can default order by "Recent" ?

lilypad/server/settings.py
```diff
@@ -14,10 +14,12 @@ class Settings(BaseSettings):
     environment: str = Field(default="production")
     port: int = Field(default=8000)
     remote_base_url: str = Field(default="https://app.lilypad.so")
-
+    api_key: str | None = None
+    project_id: str | None = None
     # GitHub OAuth settings
     github_client_id: str = Field(default="my_client_id")
     github_client_secret: str = Field(default="my_client_secret")
+
```
willbakst, line 22: nit: weird there's a space here before the comment but not above

lilypad/server/migrations/versions/0002_added_api_key_table.py
brenkao: yeah it will, they will not conflict so whichever gets merged to main first will be 0002

lilypad/server/models/api_keys.py
brenkao: yeah mostly that its never used in the client

lilypad/server/models/projects.py
brenkao: we dont currently have updated_at. do we want to add that to every table?

lilypad/server/settings.py
brenkao: added spaces above so its more consistent

lilypad/server/models/projects.py
willbakst: hmm, probably? we can do that separately from this though if it doesn't already exist

willbakst (approved):

Issue #74, opened 2024-12-19
Crash if checking isinstance on a custom LLM client


TruffleClock:
### Description

I think the `isinstance(model, AnthropicClient)` check in the code below is causing this error: `'property' object has no attribute '__name__'`

(If I always assign `formatted_sys_prompt` without the check, it works)

```
@lilypad.generation()
def fn(...):
    model = DEFAULT_LLM

    sys_prompt = construct_sys_prompt(
        ...
    )
    data_prompt = construct_data_prompt(
        ...
    )

    if isinstance(model, AnthropicClient):
         sys_prompt = cached_system_prompt(sys_prompt)

    response = model.get_completion(
        system_prompt=sys_prompt,
        prompt=data_prompt,
        temperature=0.0,
    )
    return response
```

willbakst:
I'm struggling to reproduce this. Can you share the full error message?

willbakst:
Minimal repro:

```python
from lilypad._utils import Closure


class MockAnthropicClient:
    @property
    def foo(self) -> str:
        return "a"


def fn() -> str:
    model = MockAnthropicClient()
    return model.foo


closure = Closure.from_fn(fn)
print(closure.code)

```

PR #92, opened 2025-01-08
fix: 'property' attribute error when checking isinstance on a custom LLM client

koxudaxi:
Closes: https://github.com/Mirascope/lilypad/issues/74


tests/_utils/closure/test_closure.py
```diff
@@ -405,6 +412,7 @@ def closure_with_long_function_name_that_wraps_around_fn(
         """)
 
 
+
```
willbakst, line 415: what is this extra line? same as above, need to run ruff format

lilypad/_utils/closure.py
```diff
@@ -429,6 +436,17 @@ def _collect_required_dependencies(
 
         return dependencies
 
+    @classmethod
+    def _map_child_to_parent(cls,child_to_parent: dict[ast.AST, ast.AST | None] ,node: ast.AST, parent: ast.AST | None = None) -> None:
```
willbakst, line 440: why is the formatting like this? needs ruff format

lilypad/_utils/closure.py
```diff
@@ -438,9 +456,16 @@ def collect(
...
+
+            child_to_parent = {}
+
+            self._map_child_to_parent(child_to_parent, tree)
+
             for node in ast.walk(tree):
                 if isinstance(node, ast.FunctionDef | ast.ClassDef):
-                    local_names.add(node.name)
+                    parent = child_to_parent.get(node)
+                    if isinstance(parent, ast.Module):
```
willbakst, line 467: can you explain this check and the child_to_parent mapping you create? i don't quite get what it's solving, so a comment in the code here would also be helpful, so we should include your reply/answer in the code as a comment as well

willbakst (approved):

Issue #80, opened 2024-12-19
Closure not working for some imports


brenkao:
### Description

Reproduce steps
```
from datetime import datetime

from lilypad._utils import Closure


def test_datetime() -> str:
    return datetime.now().strftime("%B %d, %Y")


print(Closure.from_fn(test_datetime))
```

Code does not contain the import
```
name='test_datetime' signature='def test_datetime() -> str: ...' code='def test_datetime() -> str:\n    return datetime.now().strftime("%B %d, %Y")\n' hash='...' dependencies={}
```

PR #93, opened 2025-01-08
fix: Closure logic for aliased imports

koxudaxi:
Closes: #80


lilypad/_utils/closure.py
```diff
@@ -334,6 +348,50 @@ def _collect_assignments_and_imports(
             self.imports.update(import_collector.imports)
             self.user_defined_imports.update(import_collector.user_defined_imports)
 
+    @classmethod
+    def _remove_conflicting_imports(cls, imports_set: set[str]) -> set[str]:
```
willbakst, line 352: what do you mean by "conflicting" imports? i'm unsure what this and the below function are doing that the final ruff formatting and fixing doesn't handle?

lilypad/_utils/closure.py
```diff
@@ -334,6 +348,72 @@ def _collect_assignments_and_imports(
...
+            import openai
+            from openai import OpenAI
+
+        If tests or docstrings expect only the "from openai import OpenAI" style,
+        we want to discard the plain "import openai" import, or vice versa.
+
+        Ruff can handle formatting and linting, but it does not always enforce
+        a single import style if there is a reason to keep "from X import Y" strictly
+        (e.g., to match exact test docstring expectations). This function explicitly
+        removes the conflicting import lines to match such requirements.
```
willbakst, line 367: if ruff doesn't resolve this, should it be resolved, or should we leave it as expected if run through ruff formatting

koxudaxi: OK,  I have deleted the method.

lilypad/_utils/closure.py
```diff
@@ -334,6 +348,26 @@ def _collect_assignments_and_imports(
             self.imports.update(import_collector.imports)
             self.user_defined_imports.update(import_collector.user_defined_imports)
 
+    @classmethod
+    def _remove_unused_aliases(
```
willbakst, line 352: does ruff not remove the unused aliases?

i don't see anything in the tests that reflect this code change?

koxudaxi: This code is more about maintaining existing behavior rather than changing it. Therefore, there are no changes to the unittest.
While we could remove it from imports by adding the F401 option to ruff, I created this method because we need to remove it from `closure.dependencies`.

willbakst: why would we need to remove it? if it's an alias but the original is used, wouldn't the dependency be the same?

koxudaxi: If I disable this method and run the test, it seems that `import openai as oai` is inserted directly under the test module.
I'll check a little more.
```python
+ import openai as oai
from mirascope.core import openai


@openai.call("gpt-4o-mini")
def decorated_fn() -> str:
    return "Hello, world!"
 ```

koxudaxi: @willbakst 
OK, I have fixed the alias logic

willbakst (approved):

PR #99, opened 2025-01-13
Optimize traces

brenkao:


lilypad/server/api/v0/traces_api.py
```diff
@@ -27,12 +28,17 @@ async def get_traces_by_project_uuid(
     project_uuid: UUID,
     session: Annotated[Session, Depends(get_session)],
 ) -> Sequence[SpanTable]:
-    """Get all traces"""
+    """Get all traces.
+
+    Child spans are not lazy loaded to avoid N+1 queries.
```
willbakst, line 33: this says "not lazy loaded" but then below the "sa_arguments" has "lazy"?

brenkao: thats why I have the comment because the default is `lazy='select'`. Whereas we are changing it to `selectin` which load the data immediately, instead of "lazy loaded"

willbakst: ah ok got it that makes sense

willbakst (approved):

Issue #85, opened 2024-12-20
Hard to review long prompts


TruffleClock:
### Description

All combined into a single line :(

<img width="1017" alt="Screenshot 2024-12-19 at 5 08 12 PM" src="https://github.com/user-attachments/assets/27bc8bc1-402f-4099-aee9-7a9507aadd05" />

willbakst:
Oh hmm. I'm going to mark this as a bug rather than an enhancement because this impacts the versioning and should retain the original docstring (rather than injects `\n`).

PR #113, opened 2025-01-17
feat: Improve Handling of Multi-Line Strings with Triple Quotes

koxudaxi:
Closes: https://github.com/Mirascope/lilypad/issues/85


lilypad/_utils/closure.py
```diff
@@ -43,11 +45,64 @@ def _is_third_party(module: ModuleType, site_packages: set[str]) -> bool:
...
+                    new_tokens.append((tok_type, triple_str))
+                else:
+                    new_tokens.append((tok_type, tok_string))
+        else:
+            new_tokens.append((tok_type, tok_string))
+
+    return tokenize.untokenize(new_tokens)
+
+
 def _clean_source_code(
```
willbakst, line 79: originally the reason that I would parse and then unparse the tree to get the source code was to remove comments from the code (because comments don't impact functionality and can change without changing the generation).

the better approach may be to instead use the original source code and clean that e.g. removing comments  (so that the original code is preserved as much as possible).

the idea is that we should try to use the original source code as closely as possible while removing anything from the code that does not impact the functionality (e.g. docstrings, comments, etc)

lilypad/_utils/closure.py
```diff
@@ -509,7 +564,9 @@ def collect(
         for code in self.source_code:
             tree = ast.parse(code)
             new_tree = rewriter.visit(tree)
```
willbakst, line 566: wrt. my other comment, we'll need to update this as well to perform the rewriting without parsing/unparsing the source code (so the original code is preserved as much as possible).

lilypad/_utils/closure.py
koxudaxi: Since converting to AST loses a lot of information, I rewrote it using tokenize. However, I think parse errors will occur with complex indentation and edge cases. I would like to confirm the direction for future maintenance.
It might be worth considering third-party libraries like LibCST that can handle additional information, rather than processing at the tokenize level.

koxudaxi: I've created a LibCST version for testing here: https://github.com/Mirascope/lilypad/pull/115
While I haven't done it yet, rewriting _QualifiedNameRewriter would probably be easier with LibCST as well. I think it would be quite difficult to reimplement _QualifiedNameRewriter using only tokenize.

willbakst: I left a comment on #115 

We should definitely rely on existing libraries where is makes sense. The PR with LibCST looks good.

Feel free to address the comment there or merge it into this PR and address it here.

koxudaxi:
@willbakst 
I made minimal changes to switch from AST to libCST implementation. 

I tried reimplementing other AST-inherited classes using libCST as well, but they didn't work properly. This is probably because the behavior of visit functions for each node is different between AST and libCST. I feel we need either fine-tuning of the implementation or a redesign of the architecture.

Since these minimal changes achieve our current goals, I think we don't need to touch other parts for now.

willbakst (approved):

PR #115, opened 2025-01-22
feat: use libcst

koxudaxi:


lilypad/_utils/closure.py
```diff
@@ -46,178 +52,147 @@ def _is_third_party(module: ModuleType, site_packages: set[str]) -> bool:
 
 
 def _convert_embedded_newlines_to_triple_quoted(code: str) -> str:
```
willbakst, line 54: why do we need this? i was assuming that we would use libcst to remove the unwanted components but leave everything else the same, so the strings should be unaffected right?

koxudaxi: You're right, it's probably unnecessary.
I just applied LibCST to the existing implementation, so some old code remains unchanged.
I think we can clean this up and remove it.

PR #126, opened 2025-01-29
feat: Run new generation versions on existing dataset for version comparison

koxudaxi:


lilypad/ee/evals/datasets.py
```diff
@@ -54,6 +55,37 @@ def __repr__(self) -> str:
...
 
+    def run(self, fn: Callable) -> None:
+        """Run a function on each row of the dataset, passing in the row data as kwargs."""
+        client = _get_client()  # Ensure client is initialized
+        current_closure = Closure.from_fn(fn)
+
+        closures: list[Closure] = [current_closure]
+
+        # Collect all closures with the same name.
+        for generation in client.get_generations_by_name(current_closure.name):
```
willbakst, line 66: why are you running every closure?

koxudaxi: I fixed. now, run only current and previous one.

willbakst: hmm, it should only run for the generation i'm currently trying to run it on, right?

if i wanted to run it on both, i would just run it on both?

koxudaxi: make sense. the function runs only the closure.

PR #136, opened 2025-02-03
feat: separate models and schemas to better handle circular dependencies

koxudaxi:


brenkao:
I don't think we need to migrate off SQLModel, ultimately it's just a convenience and we can write queries ourselves instead of using their `Relationship`. If we move to pure SQLAlchemy + Pydantic, we need to maintain a SQLAlchemy table and pydantic model which I think is more inconvenient

brenkao (approved):

PR #139, opened 2025-02-04
fix: Fix dockerfile for oxenai

koxudaxi:


Dockerfile
```diff
@@ -1,9 +1,32 @@
 # Use a Python image with uv pre-installed
 FROM ghcr.io/astral-sh/uv:python3.10-bookworm-slim
 
+# Install Cargo and build tools
+RUN apt-get update && apt-get install -y \
```
brenkao, line 5: What is rust for? Is our oxen server going to live in the same container as our backend server?

koxudaxi: Because Oxen's Python SDK depends on compiled binary files written in Rust, we need to build Oxen on the lilypad server.

willbakst: we should just be able to pip install their SDK

PR #148, opened 2025-02-06
fix: fix require_license decorator

koxudaxi:


koxudaxi:
Added skip markers to multiple failing tests.
These will need to be addressed and fixed in a separate PR.

lilypad/ee/server/require_license.py
```diff
@@ -0,0 +1,129 @@
...
+            if license_info.organization_uuid != organization_uuid:
+                raise HTTPException(
+                    status_code=status.HTTP_403_FORBIDDEN,
+                    detail="License key does not match organization",
+                )
+
+            if self.tier == Tier.ENTERPRISE and license_info.tier != Tier.ENTERPRISE:
+                raise HTTPException(
+                    status_code=status.HTTP_403_FORBIDDEN,
+                    detail="Invalid License. Contact the team to get one.",
```
brenkao, line 120: probably use Will's email here

lilypad/ee/server/api/v0/datasets_api.py
```diff
@@ -169,6 +169,7 @@ def from_metadata(
 @datasets_router.post(
     "/projects/{project_uuid}/generations/{generation_uuid}/datasets",
 )
+@require_license(tier=Tier.ENTERPRISE)
```
brenkao, line 172: same as above wrt. setting enterprise to be default

lilypad/server/api/v0/organizations_api.py
```diff
@@ -90,3 +96,42 @@ async def create_organization_invite(
...
+)
+async def update_organization(
+    organization_uuid: UUID,
+    organization_service: Annotated[OrganizationService, Depends(OrganizationService)],
+    organization_update: OrganizationUpdate,
+    user: Annotated[UserPublic, Depends(get_current_user)],
+) -> OrganizationTable:
+    """Update an organization."""
+    # Check if user is in organization
+    if not user.user_organizations or not any(
```
brenkao, line 113: we might also want to check if the user in the user_organization role is OWNER. so that MEMBERs and ADMINs cannot upgrade tier. Wdyt @willbakst

lilypad/server/api/v0/organizations_api.py
```diff
@@ -90,3 +96,42 @@ async def create_organization_invite(
...
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Only self organization can update license",
+        )
+
+    # If updating license, validate it
+    organization = organization_update.model_dump(exclude_unset=True)
+    if (
+        "license" in organization
+        and (new_license := organization["license"]) is not None
```
brenkao, line 126: For free tier, do we generate a license always, or does None assume FREE tier?

lilypad/ee/server/require_license.py
```diff
@@ -0,0 +1,129 @@
...
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail="Project not found",
+            )
+        if not project.organization_uuid:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail="Project does not belong to an organization",
+            )
+        organization_uuid = project.organization_uuid
```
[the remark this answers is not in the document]
willbakst: nit: "organization" not "project" as we should be generating a license per organization (which could have many projects, all under the same organization and thus license)

lilypad/cli/commands/license.py
```diff
@@ -0,0 +1,55 @@
...
+import requests
+import typer
+from rich import print
+
+from ...ee.validate import LicenseError, LicenseValidator
+from ...server.client import LilypadClient
+from ._utils import get_and_create_config
+
+
+def set_license_command(
```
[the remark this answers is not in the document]
willbakst: only issue is that we generate API keys for specific projects, so I'm not sure that's quite right?

maybe it makes the most sense to remove this and have the user (with OWNER or ADMIN privileges) set the license key in the UI (like we do for provider API keys)?

lilypad/ee/server/api/v0/annotations_api.py
```diff
@@ -10,16 +10,17 @@
...
 
-validate_license()
 annotations_router = APIRouter()
 
 
 @annotations_router.post(
     "/projects/{project_uuid}/annotations",
     response_model=Sequence[AnnotationPublic],
 )
+@require_license(tier=Tier.ENTERPRISE)
```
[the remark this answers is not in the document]
willbakst: personally I think there should be no default value for `tier` so that it's always required and immediately clear what tier and endpoint it

lilypad/ee/server/api/v0/datasets_api.py
willbakst: see my other comment

lilypad/ee/server/require_license.py
willbakst: same for the first statement as well. although maybe use `support@mirascope.com`

lilypad/server/api/v0/main.py
```diff
@@ -30,11 +36,15 @@
 api.include_router(traces_router)
 api.include_router(auth_router)
 api.include_router(users_router)
-api.include_router(organization_invites_router)
+api.include_router(organization_router)
 api.include_router(annotations_router)
```
willbakst, line 40: remove these since they are part of ee now?

lilypad/server/api/v0/organizations_api.py
willbakst: yeah for now only the highest member role (OWNER) should be able to set the tier

lilypad/server/api/v0/organizations_api.py
willbakst: it looks like if there is a license with tier "FREE" we return None so that both are the same

lilypad/cli/commands/license.py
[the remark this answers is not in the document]
willbakst: removing this for now

willbakst (approved):

Issue #176, opened 2025-02-09
Saving spans with nested generations fails.


brenkao:
### Description

Example problem right now
```
# First request
Span ID:  dfbf68f27b72cde3
Parent Span ID:  f8916186d64d57aa
Span ID:  f8916186d64d57aa
Parent Span ID:  69bf1353e9f5221e
```

Note that ` 69bf1353e9f5221e` does not exist in the first request
```
# Second request
Span ID:  3976e9aa6090ac23
Parent Span ID:  7ccad9a13942f31f
Span ID:  7ccad9a13942f31f
Parent Span ID:  d16de3bde9ed8f60
...
Span ID:  69bf1353e9f5221e
Parent Span ID:  None
```

It shows up in the second request because of the nested generations.

Nested generations are common when dealing with Agent Loops.

Option 1: Easiest solution but can cause problems is to remove the foreign key on `parent_span_id`
Option2: Keep the foreign key on `parent_span_id` on the `spans` table, but defer sending the spans until the entire loop is finished.

### Python, Lillypad & OS Versions, related packages (not required)

```TOML

```

willbakst:
I would think waiting to send the spans until the entire loop is finished is the right path. I believe this would also resolve #135 since they would only output at the end?

Only thing that comes to mind downstream would be to ensure that we properly handle errors. Might be best to send error spans that still trace against the generation but are marked with their error. This would help with debugging and also unsure the success of children generations that succeed would still be traced even if we delay until the end.

PR #183, opened 2025-02-10
Fix: Adjust flush timing for nested generations using outermost_lock_context

koxudaxi:
Fixes: https://github.com/Mirascope/lilypad/issues/176

Overview:
Fix the issue where spans in nested generations are flushed prematurely, causing foreign key constraint failures due to missing parent spans.

Changes:
- Modified the flush logic so that only the outermost generation performs a flush, while inner generations do not trigger a flush.
- Renamed and updated the context manager from `manual_flush_context()` to outermost_lock_context(). When enabled for the outermost generation (`enable_lock=True`), `outermost_lock_context()` acquires the BatchSpanProcessor's condition lock (via "with processor.condition:") to synchronize flush operations. This ensures that flush occurs only once at the end of the outermost generation.
- Updated the generation() decorator to use outermost_lock_context(is_outermost) appropriately, ensuring that flush is executed only for the outermost generation.

Effect:
This change ensures that all spans are sent only after the entire nested generation has completed, preserving the parent-child relationships and avoiding foreign key constraint issues.

### DIFF
--- a/lilypad/generations.py
+++ b/lilypad/generations.py
@@ -4,7 +4,7 @@ import inspect
 import json
 from collections.abc import Callable, Coroutine, Generator
 from contextlib import contextmanager
-from contextvars import ContextVar
+from contextvars import ContextVar, Token
 from functools import wraps
 from typing import Any, ParamSpec, Protocol, TypeVar, overload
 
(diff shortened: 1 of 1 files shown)


lilypad/generations.py
```diff
@@ -49,6 +51,46 @@ def __call__(
...
+    to export all buffered spans. For inner generations (flush=False), no change is made.
+    """
+    if not flush:
+        yield
+        return
+    processor = _get_batch_span_processor()
+    old_delay = None
+    if processor is not None:
+        old_delay = processor.schedule_delay_millis
+        processor.schedule_delay_millis = 10**12  # Effectively disable auto flush.
```
brenkao, line 85: can we use inf here or does that not work?

brenkao (approved):

lilypad/generations.py
koxudaxi: OK, I changed the strategy.
the context manager get the lock file for worker in `BatchProcessor`.
and, the worker flush after exit the outermost generation.

PR #196, opened 2025-02-17
feat: migrate oxen to oxenfs

koxudaxi:
Do not merge because the current implementation does not create the directory


willbakst:
We've removed Oxen for the time being while we iron out the annotations interface and wait for updates on their end wrt. how we will more easily interact with their system down the line. Closing this until future work on this is done.

Issue #188, opened 2025-02-10
Prompt fails to run outside of a generation


willbakst:
### Description

I tried running essentially this:

```python
import lilypad

@lilypad.prompt()
def my_prompt(): ...

prompt = my_prompt()
print(prompt.messages("openai"))
```

It fails saying there is no active version, but the default is selected in the UI. When used inside a generation it works.

### Python, Lillypad & OS Versions, related packages (not required)

```TOML
python-lilypad[openai]==0.0.15
```

willbakst:
Most likely happening here: [https://github.com/Mirascope/lilypad/blob/596859082bcc607c27e027e8a8d80d1c18ad8ca9/lilypad/server/client.py#L263](https://github.com/Mirascope/lilypad/blob/596859082bcc607c27e027e8a8d80d1c18ad8ca9/lilypad/server/client.py#L263)

willbakst:
I'm closing this as we are removing managed prompts in favor of upcoming managed generations where this won't be a problem.

PR #198, opened 2025-02-18
fix: Allow Prompts to Run Outside of Generation Context by Using Default Version

koxudaxi:
Fixes: https://github.com/Mirascope/lilypad/issues/188


lilypad/server/client.py
```diff
@@ -271,41 +272,58 @@ def get_prompt_active_version(
...
+                    return prompts[0]
         try:
             return self._request(
                 "GET",
                 f"v0/projects/{self.project_uuid}/prompts/hash/{closure.hash}/active",
                 response_model=PromptPublic,
             )
         except NotFoundError:
-            return None
+            prompts = self._request(
```
brenkao, line 303: Do we want to create a empty prompt if we dont find one? Shouldn't we print a link to the UI to create a prompt? @willbakst

willbakst: I think we can print a UI link. Given recent discussion around managed prompt -> managed generation switch, I'm not sure how much this matter anymore.

koxudaxi:
@willbakst @brenkao 
Given the significant changes involved in the managed prompt -> managed generation switch, what's your preference - should we merge this ticket and PR #199 first, or should we close them and focus on the switch implementation?

willbakst:
> @willbakst @brenkao Given the significant changes involved in the managed prompt -> managed generation switch, what's your preference - should we merge this ticket and PR #199 first, or should we close them and focus on the switch implementation?

let's merge these first and then focus on the migration

lilypad/server/client.py
koxudaxi: @brenkao 
> Do we want to create a empty prompt if we dont find one?

Do you have any recommended idea?

willbakst: Hmm, I think for now if there is no prompt found when trying to run it, we should just raise an error that tells them to create one with a link to the prompt creation UI

koxudaxi: OK, I fixed to raise an exception for messages containing URLs when a record is not found.

Issue #201, opened 2025-02-18
Playground Updates


koxudaxi:
Migrate the current "Managed Generation" playground into its own "Playground" tab that you can just use as a standard playground.

- [X] Refactor "Managed Generations" into a "Playground" tab. The code that we generate to run should be the unwrapped Lilypad + Mirascope code so that versioning on the closure is sufficient.
- [X] When you click on the playground, it should open to the playground with no versioned function selected. If you try to hit "Run" it should notify you (highlight red?) that no function is selected. When you go to select the function you should be able to either (1) select an existing function or (2) create a new one (which is just entering a name where the version will be actually created when it runs). We might want to move the selected function version dropdown and the "Run" button to be next to each other. We should remove the "Save" button. You have to run it to save it because "saving" really means running and versioning the function.
- [X] There should be an additional select for hot swapping versions of an attached function.
- [X] Side-by-side playground sessions

koxudaxi:
@willbakst @brenkao 
What do you think of this idea? pelase leave a comment. 

**Implementation Details for Transition from Managed Prompts to Managed Generations**

1. **Review of Hash and Signature Calculation**  
   - **Rename Column:**  
     We will change the current `hash` column in the PromptTable to `template_hash`. This will show that the hash is computed only from the prompt template.  
   - **Add New Column (`closure_hash`):**  
     We will add a new column called `closure_hash` (a string of about 64 characters if using SHA-256). This column is computed from the full closure, including the normalized signature, code, and argument definitions. This change will update the hash when the prompt changes in a meaningful way.  
   - **Client-Side Signature Generation:**  
     The client will use `construct_function` to generate a normalized prompt signature. The client will then calculate both `template_hash` and `closure_hash` and send these values to the API. The server will use the given signature directly without recalculating it. This helps to keep the client and server in agreement.

2. **API Endpoint and Business Logic Updates**  
   - **Endpoint Revision:**  
     We will modify the existing `/prompts` endpoints to work with the new Managed Generations approach. This includes changing the logic for prompt creation, update, and search. The new logic will use the client-provided signature and new hash values for proper version tracking.  
   - **Generation Linkage:**  
     In the GenerationTable, we may add a new column such as `prompt_closure_hash` (or `prompt_version`). This column will save a snapshot of the prompt’s version (for example, its `closure_hash` or `version_num`) at the time a generation is created. This change helps downstream systems know which version of the prompt was used.

3. **Database Schema Changes**  
   - **PromptTable:**  
[...]

PR #207, opened 2025-02-19
feat: reimplement managed generations from old synced prompts

koxudaxi:
Related Issues: https://github.com/Mirascope/lilypad/issues/201


lilypad/_utils/closure.py
```diff
@@ -38,10 +39,17 @@ class DependencyInfo(TypedDict):
...
+    If the function is defined locally, return the name after '<locals>.'; otherwise,
+    return the last non-empty part after splitting by '.'.
+    """
+    qualified_name = fn.__qualname__
+    if "<locals>." in qualified_name:
+        # For local functions, return the part after "<locals>."
+        return qualified_name.split("<locals>.")[-1]
+    else:
+        parts = [part for part in qualified_name.split(".") if part]
+        return parts[-1] if parts else qualified_name
```
koxudaxi, line 52: While implementing this, I noticed that this is essentially a simple qualified_name. Should we consider storing qualified_names (which may include module names, etc.) as a separate field from name in both Closure and DB for future extensibility?

lilypad/_utils/closure.py
```diff
@@ -842,5 +849,90 @@ def run(self, *args: Any, **kwargs: Any) -> Any:
...
+        Args:
+            module_name: Optional name for the module.
+
+        Returns:
+            The created module.
+        """
+        import uuid
+
+        unique_suffix = uuid.uuid4().hex
+        name = module_name or f"dynamic_module_{self.hash[:8]}_{unique_suffix}"
```
koxudaxi, line 869: Avoid leaks by name

lilypad/messages.py
```diff
@@ -1,99 +1,31 @@
...
-        mcls,
-        name: str,
-        bases: tuple[type, ...],
-        namespace: dict[str, Any],
-        **kwargs: Any,
-    ) -> type:
-        cls = super().__new__(mcls, name, bases, namespace)
-        cls.__abstractmethods__ = frozenset()
-        return cls
+class Message(CallResponse):
```
koxudaxi, line 11: Since we've already completed a practical CallResponse in mirascope, I wrapped it in a simple way. I haven't checked the requirements in detail though - please let me know if this differs from what you had in mind.

lilypad/server/client.py
```diff
@@ -249,44 +258,133 @@ def get_or_create_generation_version(
...
+            )
+        generations = self._request(
+            "GET",
+            f"v0/projects/{self.project_uuid}/generations/name/{closure.name}",
+            response_model=list[GenerationPublic],
+        )
+
+        for generation in generations:
+            if generation.version_num == forced_version_num:
+                return generation
```
koxudaxi, line 290: We will likely need to review the API requests and queries in the future, taking performance considerations into account.

lilypad/server/client.py
```diff
@@ -249,44 +258,133 @@ def get_or_create_generation_version(
...
+        generations = self._request(
+            "GET",
+            f"v0/projects/{self.project_uuid}/generations/name/{closure.name}",
+            response_model=list[GenerationPublic],
+        )
+        # Check only args count for now
+        args_count = len(args_types)
+        for generation in generations:
+            if len(generation.arg_types) == args_count:
+                return generation
```
koxudaxi, line 319: At least we can avoid mismatched argument counts. Should we implement more detailed checks?

lilypad/server/models/generations.py
```diff
@@ -51,6 +52,12 @@ class _GenerationBase(SQLModel):
     arg_types: dict[str, str] = Field(sa_column=get_json_column(), default_factory=dict)
     archived: datetime | None = Field(default=None, index=True)
     custom_id: str | None = Field(default=None, index=True)
+    prompt_template: str | None = Field(default=None)
+
+    # There may be objects in the dictionary that cannot be serialized?
+    call_params: dict[str, Any] = Field(
+        sa_column=get_json_column(), default_factory=dict
+    )
```
koxudaxi, line 60: While this should work fine in most cases, we may need to consider in the future that some LLM providers might be using Python objects like `pydantic.BaseModel` as call_params.

lilypad/_utils/closure.py
brenkao: What do we get from saving qualified_names?

koxudaxi: The UI details would be helpful to output. When using this feature in a large product, I thought it would be difficult to identify the desired generation to edit just by function name.

lilypad/server/client.py
brenkao: We can create a get closure version number endpoint

lilypad/server/models/generations.py
brenkao: Ideally we would use JsonableType but that caused some issues here, so Any is fine temporarily.

lilypad/server/models/tools.py
```diff
@@ -0,0 +1,56 @@
...
+
+class ToolTable(_ToolBase, BaseOrganizationSQLModel, table=True):
+    """Table for response models."""
+
+    __tablename__ = TOOL_TABLE_NAME  # pyright: ignore [reportAssignmentType]
+
+    project_uuid: UUID | None = Field(
+        default=None, foreign_key=f"{PROJECT_TABLE_NAME}.uuid", ondelete="CASCADE"
+    )
+    span_uuid: UUID | None = Field(
```
brenkao, line 48: We probably dont need a connection from span to tools, because we already have the connection between span and generations, which has the tools.

lilypad/server/models/tools.py
```diff
@@ -0,0 +1,56 @@
...
+        default=None, foreign_key=f"{SPAN_TABLE_NAME}.uuid", ondelete="CASCADE"
+    )
+    generation_uuid: UUID | None = Field(
+        default=None, foreign_key=f"{GENERATION_TABLE_NAME}.uuid", ondelete="CASCADE"
+    )
+    project: "ProjectTable" = Relationship(back_populates="tools")
+    span: "SpanTable" = Relationship(back_populates="tools")
```
brenkao, line 55: see above

lilypad/server/client.py
koxudaxi: OK, I have created it

willbakst:
I changed the PR to point to a new `managed-generations` branch so that we don't accidentally break things in `main`.

Once this is merged into the new feature branch, @brenkao can take care of the client updates / make sure everything works before we release the feature.

lilypad/_utils/closure.py
willbakst: can you elaborate?

lilypad/messages.py
willbakst: I think this is good, originally we didn't have `CallResponse` but now that we do we should utilize it

lilypad/server/client.py
willbakst: for managed generations, we should require that there is only ever one valid signature. If the user want to remove/add/change arguments, they'll have to go through some flow in the UI as well to ensure that we know what the pending -> deployed signature state is.

lilypad/server/client.py
```diff
@@ -249,44 +258,131 @@ def get_or_create_generation_version(
...
+        raise LilypadNotFoundError(
+            f"Generation args_types '{args_types}' not found for signature {closure.signature}"
+        )
+
     def get_prompt_active_version(
```
willbakst, line 322: shouldn't we be removing all the prompt stuff since we'll be putting it all under generations when managed? The "active" version is just that generations version? If they change the template, that would be a new version (even if the signature doesn't change)

lilypad/_utils/closure.py
```diff
@@ -842,5 +849,90 @@ def run(self, *args: Any, **kwargs: Any) -> Any:
...
+            obj = getattr(module, self.name)
+        except AttributeError as e:
+            raise AttributeError(f"Object {self.name} not found in module") from e
+        finally:
+            # Clean up: Remove the module from sys.modules to avoid interference between requests.
+            sys.modules.pop(module.__name__, None)
+        return obj
+
+    @classmethod
+    def from_code(
```
willbakst, line 914: where do we use this method? it looks like it only appears in tests? We use the `from_fn` method generally, what is this method doing?

koxudaxi: The PR uses the method to re-build `response_model` and `tools` from DB
https://github.com/Mirascope/lilypad/blob/0881bdf18edd1985bc2087a9167023df28916d99/lilypad/generations.py#L304-L313

lilypad/server/client.py
koxudaxi: I will remove all prompt stuff

lilypad/_utils/closure.py
koxudaxi: @willbakst 
Maybe I misunderstood? 
I was thinking that tools would be managed in the DB just like prompts when they are fully managed, but were you assuming that the code part of tools (function body) would be stored in the source code and only the interface would be stored in DB?

lilypad/_utils/closure.py
koxudaxi: When generating objects from source code on the server, using a fixed module name creates a risk of object collision during object creation. By using UUID, I tried to avoid this problem.

lilypad/_utils/closure.py
willbakst: yeah, I envision the implementation of the tool (e.g. searching the web) existing as code because that is what makes the most sense to me.

They would then be included in a generation's version when used in a generation. I imagine (for now at least) that we only have Generations and everything else is simply part of that (whether or not it's managed)

lilypad/_utils/closure.py
willbakst: does this change with my other comment?

lilypad/server/client.py
koxudaxi: OK, I have fixed it.

lilypad/_utils/closure.py
koxudaxi: What about ResponseModel?
I think when we talked quite a while ago, the plan was to build pydantic models from jsonschema.
If it's the same approach, we would need to build from code, so similar code would be necessary.
How will we handle ResponseModel?
Will we handle it in source code like Tools, or in the DB as jsonschema?
Which approach are we taking?

lilypad/_utils/closure.py
koxudaxi: I commented. it depends on ResponseModel implementation

lilypad/server/client.py
koxudaxi: I have deleted it.

lilypad/_utils/closure.py
willbakst: So there won't be a separate handling of response models, they will be handled as part of managed generations. If you're not using a managed generation, then the response model will need to live in the code. Same thing for tools. Previously we were treating response models and tools the same way as prompts (separately from generations), but now we're merging everything under the single hood of Managed Generations.

If you're using a managed generation and the return type is a response model, then you should be able to add things like descriptions etc. for fields in the UI (the same way we can change the prompt or the call params). Making a change here would change the version of that managed generation. If you want to add a new field, you would have to go through the same steps as if adding a new argument (or changing the return type) of a managed generation. The same would be true for tools (e.g. updating the description, adding field descriptions, templating the return value into a string output, etc.)

At a high level, everything falls under the same umbrella, which is the managed generation's signature. Using a response model or adding tools change the signature of the managed generation, and updating the response model schema or tool schema / signature counts as an update to the signature.

In the short term, we can require that any and all signature updates happen in the UI followed by the deployment flow of that new signature. We can think through letting updates happen in code for managed generations later once the core flow is fleshed out.

Does this make sense?

willbakst: It's worth noting that we need to be able to run managed generations the same on our server as the user runs it locally based only on the signature of the managed generation. This also means that in the short term we should avoid tool use for managed generations given that tools can contain arbitrary user code. For now, users will have to use standard generations (not managed) to use tools.

lilypad/_utils/closure.py
willbakst: Got it. I responded to that thread. I will resolve this and reopen if necessary.

lilypad/_utils/closure.py
koxudaxi: @willbakst

I have already implemented code that rebuilds tools and response_model from the DB using the `Closure.from_code` method, but I think we need to further clarify this implementation approach.

Based on your latest direction, is my understanding correct:

1. Integrated Architecture:
   - Manage everything (prompts, ResponseModel, tools) as part of Managed Generations
   - For non-Managed Generation, implement everything in the code

2. For Managed Generation:
   - To make it work on both local and server, we need to store code representations of tools and ResponseModel in the DB
   - This requires serialization/deserialization (I understand from the context of removing `Closure.from_code` that building from JsonSchema would be correct)

3. About Tools and Signatures:
   - Signature changes (including structural changes to tools/ResponseModel) trigger version changes
   - Version changes are made through the UI, following the deployment flow
   - In the short term, tool use in Managed Generations should be avoided due to arbitrary user code in tools

Questions:
1. Is this understanding correct?
2. In the long term, how do you plan to handle tools in Managed Generation? Allow managing tool code in the UI? Or always through source code management?

I will modify the implementation to prohibit using tools in managed generations for the short term, as per your guidance.

willbakst: I think you're understanding correctly.

The closure computation should make it possible already to run the tool code, so it's really just a matter of implementing a secure way of doing that (i.e. sandbox etc.) before we turn it on. Once we can run it securely, it should be as simple as adding the tool's method body from the user's perspective. Since updating the tool's body updates the closure but not the signature, it would be safe for a user to update the tool's body and create a new managed generation version by running the code with the updated body and matching signature.

Also, I'm not even sure if we need to code -> UI sync, so until that's clear we'll do UI -> code only as discussed.

koxudaxi: I agree with your approach. I'll proceed with the implementation as discussed.

koxudaxi:
OK, I updated the generation.
Could you please review it again?

lilypad/response_models.py
```diff
@@ -68,7 +67,9 @@ def _json_schema_to_python_type(schema: dict[str, Any], name: str = "Model") ->
         return Any
 
 
-def _create_model_from_json_schema(json_schema: dict[str, Any]) -> type[BaseModel]:
+def _create_model_from_json_schema(
```
willbakst, line 70: We shouldn't be fully creating the model from the JSON schema, but instead populating an existing model. This may still require using the `create_model` method, but we should be setting the `base` to the given, existing model definition, right? I imagine something like:

```python
def _populate_model_info_from_json_schema(
    model_type: type[BaseModelT], json_schema: dict[str, Any]
) -> type[BaseModelT]
```

Note the function name change in my example to `_populate_model_info_from_json_schema`, which I think is more accurate to what the method should be doing?

lilypad/server/api/v0/generations_api.py
```diff
@@ -56,6 +56,22 @@ async def get_generations_by_name(
     return generation_service.find_generations_by_name(project_uuid, generation_name)
 
 
+@generations_router.get(
+    "/projects/{project_uuid}/generations/name/{generation_name}/version/{version_num}",
+    response_model=GenerationPublic,
+)
+async def get_generations_by_version(
```
willbakst, line 63: the name is `get_generations_by_version` plural, and the function body returns `find_generations_by_version` plural, but the return type is `GenerationTable` singular, so there is some weird naming happening here that needs to be fixed

lilypad/server/models/generations.py
```diff
@@ -51,12 +47,18 @@ class _GenerationBase(SQLModel):
     arg_types: dict[str, str] = Field(sa_column=get_json_column(), default_factory=dict)
     archived: datetime | None = Field(default=None, index=True)
     custom_id: str | None = Field(default=None, index=True)
+    prompt_template: str | None = Field(default=None)
+
+    # There may be objects in the dictionary that cannot be serialized?
+    call_params: dict[str, Any] = Field(
```
willbakst, line 53: we should require that `call_params` have the Mirascope `CommonCallParams` type since managed generations are provider-agnostic, right?

we should also then be able to ensure that all fields are serializable

lilypad/server/models/generations.py
```diff
@@ -70,7 +72,7 @@ class GenerationTable(_GenerationBase, BaseOrganizationSQLModel, table=True):
     spans: list["SpanTable"] = Relationship(
         back_populates="generation", cascade_delete=True
     )
-    prompt: Optional["PromptTable"] = Relationship(back_populates="generations")
+
     response_model: Optional["ResponseModelTable"] = Relationship(
```
willbakst, line 76: as discussed, since everything will be under generations, shouldn't we remove `ResponseModelTable` and `ToolTable` like we are doing for `PromptTable` and push everything inside of the `GenerationTable` here?

For example, `response_model` could just be the JSON schema for the model, and `tools` could just be a list of pydantic models representing the tools?

lilypad/server/models/projects.py
```diff
@@ -32,9 +32,6 @@ class ProjectTable(_ProjectBase, BaseOrganizationSQLModel, table=True):
     generations: list["GenerationTable"] = Relationship(
         back_populates="project", cascade_delete=True
     )
-    prompts: list["PromptTable"] = Relationship(
-        back_populates="project", cascade_delete=True
-    )
     response_models: list["ResponseModelTable"] = Relationship(
```
willbakst, line 35: see other comment re: pushing this inside of generations

lilypad/server/models/spans.py
```diff
@@ -45,9 +51,6 @@ class SpanBase(SQLModel):
     generation_uuid: UUID | None = Field(
         default=None, foreign_key=f"{GENERATION_TABLE_NAME}.uuid", ondelete="CASCADE"
     )
-    prompt_uuid: UUID | None = Field(
-        default=None, foreign_key=f"{PROMPT_TABLE_NAME}.uuid", ondelete="CASCADE"
-    )
     response_model_uuid: UUID | None = Field(
```
willbakst, line 54: see other comment re: pushing this inside of generations (which means we'll completely remove this)

lilypad/server/models/spans.py
```diff
@@ -80,7 +83,6 @@ class SpanTable(SpanBase, BaseOrganizationSQLModel, table=True):
         default=None, foreign_key=f"{PROJECT_TABLE_NAME}.uuid", ondelete="CASCADE"
     )
     generation: Optional["GenerationTable"] = Relationship(back_populates="spans")
-    prompt: Optional["PromptTable"] = Relationship(back_populates="spans")
     response_model: Optional["ResponseModelTable"] = Relationship(
```
willbakst, line 86: same here. i'm not going to comment on any more instances of this after this comment, so just make sure to do a sweep and i'll do a second sweep check during final reviews

lilypad/server/models/tools.py
```diff
@@ -0,0 +1,50 @@
+"""Tool for the database."""
```
willbakst, line 1: it's worth noting that while removing this (as per my other comment / discussions) means that every generation will maintain it's own version of the tools it's using, this does not prevent users from re-using tools across generations since the code itself will live as code. It only means that generation specific things (like descriptions, templating, etc.) can be made generation specific. Of course, they could also remain generation agnostic by implementing even things like descriptions in code.

mostly just noting here that this doesn't remove the ability to re-use things. same is true for response models.

feel free to resolve this comment, it's just a note

lilypad/generations.py
```diff
@@ -3,38 +3,117 @@
...
+    def __call__(self, fn: Callable[_P, _R]) -> SyncGenerationFunction[_P, _R]: ...
+
+    def __call__(
+        self, fn: Callable[_P, _R] | Callable[_P, Coroutine[Any, Any, _R]]
+    ) -> SyncGenerationFunction[_P, _R] | AsyncGenerationFunction[_P, _R]:
+        """Protocol `call` definition for `generation` decorator return type."""
+        ...
+
+
+class GenerationVersioningManagedDecorator(Protocol):
```
willbakst, line 94: nit: should this be `GenerationVersioningDecorator` and `ManagedGenerationVersioningDecorator` so the naming matches the Sync/Async naming above?

lilypad/generations.py
```diff
@@ -204,7 +283,76 @@ def inner(*args: _P.args, **kwargs: _P.kwargs) -> _R:
...
+    call_params = generation_public.call_params
+    if generation_public.tools:
+        pass
+        # TODO: tools are not yet supported in managed mode
+        # Skip tools in managed mode to avoid unexpected behavior
+    elif response_model := generation_public.response_model:
+        call_params["response_model"] = _create_model_from_json_schema(
+            response_model.schema_data, response_model.name
+        )
+    mirascope_call = llm.call(**call_params)(mirascope_prompt)
```
willbakst, line 312: this seems slightly incorrect? shouldn't it be `llm.call(call_params=call_params)` and any other info (e.g. `response_model` should be set separately? unless `generation_public.call_params` contains the common call params as a property called `call_params`, but if that's the case we should change it because it isn't clear. I think we should store things like provider, model, etc. separately from call parameters so it's more clear

lilypad/generations.py
```diff
@@ -204,7 +283,76 @@ def inner(*args: _P.args, **kwargs: _P.kwargs) -> _R:
...
+    @wraps(mirascope_call)
+    def inner(
+        *args: _P.args, **kwargs: _P.kwargs
+    ) -> Message | Coroutine[Any, Any, Message | Stream] | Stream:
+        result = mirascope_call(*args, **kwargs)
+        if fn_is_async(mirascope_call):
+
+            async def async_wrapper() -> Message | Stream:
+                final = await result
+                if isinstance(final, CallResponse):
```
willbakst, line 323: what about when the output type is a response model or iterable of response models?

lilypad/generations.py
```diff
@@ -204,7 +283,76 @@ def inner(*args: _P.args, **kwargs: _P.kwargs) -> _R:
...
+
+_ArgTypes: typing.TypeAlias = dict[str, str]
+
+
+@overload
+def generation(  # pyright: ignore [reportOverlappingOverload]
+    custom_id: str | None = ..., managed: Literal[True] = ...
+) -> GenerationVersioningManagedDecorator: ...
+@overload
+def generation(
```
willbakst, line 350: nit: add newlines between the overloads for better readability

lilypad/generations.py
```diff
@@ -204,7 +283,76 @@ def inner(*args: _P.args, **kwargs: _P.kwargs) -> _R:
...
+
+    return inner  # pyright: ignore [reportReturnType]
+
+
+_ArgTypes: typing.TypeAlias = dict[str, str]
+
+
+@overload
+def generation(  # pyright: ignore [reportOverlappingOverload]
+    custom_id: str | None = ..., managed: Literal[True] = ...
```
willbakst, line 347: we should implement a `Generation` class that contains all of the information for a generation and include an argument that enables returning the `Generation` class instead of the original response. This would make it possible to do things down the line like forward annotations for a specific generation / trace to the user's user. For example, they may want to call the generation, handle the response, and then also let their user annotate the trace, which is not currently possible.

I'm thinking we add an argument e.g. `mode: GenerationMode = "no-wrap"` with an option "wrap" that forces the `Generation` return type. We would then make it generic on the original type (i.e. `Generation[Message]`) where the user can access the message / original response through something like a `output` field or something.

WDYT? I'm not set on the naming, so ideas are welcome here

lilypad/_utils/closure.py
```diff
@@ -842,5 +849,47 @@ def run(self, *args: Any, **kwargs: Any) -> Any:
         finally:
             tmp_path.unlink()
 
+    def create_module(
```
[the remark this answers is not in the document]
koxudaxi: I forget to delete it.

lilypad/call_response_chunk.py
```diff
@@ -0,0 +1,7 @@
+"""The lilypad `CallResponseChunk` class."""
+
+from mirascope.llm.call_response_chunk import CallResponseChunk as _CallResponseChunk
+
+
+class CallResponseChunk(_CallResponseChunk):
```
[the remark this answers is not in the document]
koxudaxi: I agree with you I will select `Message and MessageChunk`

lilypad/response_models.py
```diff
@@ -183,6 +184,8 @@ def get_response_model_version(
         ) -> ResponseModelPublic:
             lilypad_client = LilypadClient()
             """Get or create the active version of this response model."""
+            from lilypad.generations import current_generation
```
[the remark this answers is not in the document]
koxudaxi: Because of an error caused by circular references during import.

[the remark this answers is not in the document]
willbakst: I think it would be better to resolve the circular import rather than lazy load if possible

lilypad/response_models.py
koxudaxi: I deleted the logic

lilypad/generations.py
koxudaxi: we don't need care response _model pattern.

lilypad/generations.py
koxudaxi: @willbakst 
Since we're not supporting response_model and tools for now, can we postpone this feature for later? Or should we consider it now?

lilypad/generations.py
willbakst: Yeah since we're removing it we can ignore this

lilypad/generations.py
willbakst: I created a new issue and will resolve this

lilypad/generations.py
```diff
@@ -204,7 +283,79 @@ def inner(*args: _P.args, **kwargs: _P.kwargs) -> _R:
     return decorator
 
 
-def generation(custom_id: str | None = None) -> GenerationDecorator:
+@overload
+def _build_mirascope_call(  # pyright: ignore [reportOverlappingOverload]
+    generation_public: GenerationPublic, fn: Callable[_P, Coroutine[Any, Any, _R]]
+) -> Callable[_P, Coroutine[Any, Any, Message | Stream]]: ...
+@overload
```
willbakst, line 290: nit: also add newlines here like mentioned elsewhere

willbakst (approved):

Issue #213, opened 2025-02-19
Stats Analysis (e.g. cost, tokens, latency, etc.)


willbakst:
### Description

We currently track the cost/tokens/latency of individual API calls, but we don't have an easy way to view overall stats / compare across versions. This makes analysis difficult.

For example, we may want to see if version 2 is more expensive than version 1 of a generation to determine if the cost difference is worth the performance change.

There are a few things we should consider adding here:

- [X] Stats for each generation in aggregate as part of the trace.
  * For example, a parent generation that makes two calls to OpenAI should show the total cost/tokens across both calls as well as the total time of the parent generation (which is different from the sum of the individual OpenAI calls' time because there could be additional work done in the parent generation that adds to the total time).
- [X] Overall stats for a generation across many traces.
  * For example, we may want to view average cost/latency per call by day / month, total cost, total tokens, etc. I think some form of charting here would be good.
- [X] Overall project stats.
  * For example, we may want to view total cost per month for a project.

We will likely want to store some form of metric data so that we don't have to compute this information for every request.

PR #226, opened 2025-02-22
feat: bubble up metrics to parent spans and display them in client

brenkao:
First part of #213 

This is the backend portion


lilypad/server/_utils/spans.py
```diff
@@ -398,6 +398,13 @@ def calculate_cost(
     https://openai.com/pricing
 
     Model                   Input               Output
```
willbakst, line 400: for now this is fine, but i feel like we should think through some form of merging of our cost calculation stuff since we have this both here and in mirascope

i'm thinking we might want to have a separate package so that we can update the pricing calculations without additional deps (i.e. someone who doesn't want to upgrade their mirascope version could still upgrade their cost calculation version to get the new stuff)

Issue #200, opened 2025-02-18
Define Expected Behavior for Class Context in @lilypad.generation()


koxudaxi:
## Current Situation
The `@lilypad.generation()` decorator currently loses class context when used with class methods. Example:

```python
class Chatbot:
    @lilypad.generation()
    def run(self) -> str:
        return "Hello, world!"
```

Gets transformed to:

```python
@lilypad.generation()
def run(self) -> str:
    return "Hello, world!"
```

## Discussion Points

Before implementing any fixes, we need to agree on the expected behavior:

1. Should the decorator support class methods at all?
   - What are the use cases for having generation inside a class?
   - Would it be better to keep generation functions separate from classes?

2. If we support class methods:
   - What class attributes should be accessible?
   - How should we handle initialization parameters?
   - What should happen with inherited methods and attributes?

3. Alternative approaches to consider:
   - Moving generation logic outside of classes
   - Creating a different decorator specifically for class methods
   - Passing necessary context as explicit parameters

## Examples of Potential Use Cases

```python
class Chatbot:
    def __init__(self, system_prompt: str):
        self.system_prompt = system_prompt
    
    @lilypad.generation()
    def run(self, user_input: str) -> str:
        # Should this have access to self.system_prompt?
        # How should we handle the system prompt in generation?
        pass
```

## Decisions Needed

The following decisions must be made before proceeding with implementation:

1. Target Use Case Definition
   - Identify primary use cases for class-based generation
   - Determine if class-based implementation is necessary for these use cases

2. Parameter Handling Strategy
   - Choose between class context capture and explicit parameter passing
   - Define the scope of accessible class attributes if context capture is chosen

These architectural decisions will guide the implementation approach and API design.

## Related Isses
- https://github.com/Mirascope/lilypad/issues/191

### Python, Lillypad & OS Versions, related packages (not required)

```TOML

```

PR #227, opened 2025-02-22
fix: fix closure with class method

koxudaxi:
Fixes: https://github.com/Mirascope/lilypad/issues/200

### DIFF
--- a/lilypad/_utils/closure.py
+++ b/lilypad/_utils/closure.py
@@ -14,7 +14,7 @@ import subprocess
 import sys
 import tempfile
 from collections.abc import Callable
-from functools import cached_property, lru_cache
+from functools import lru_cache
 from pathlib import Path
 from textwrap import dedent
 from types import ModuleType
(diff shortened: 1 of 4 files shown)


lilypad/_utils/closure.py
```diff
@@ -498,6 +498,39 @@ def leave_Name(self, original_node: cst.Name, updated_node: cst.Name) -> cst.Nam
...
 
+def _get_class_from_unbound_method(method: Callable[..., Any]) -> type | None:
+    qualname = method.__qualname__
+    parts = qualname.split(".")
+    if len(parts) < 2:
+        return None
+    class_qualname = ".".join(parts[:-1])
+    import gc
+
+    for obj in gc.get_objects():
```
koxudaxi, line 509: It's diffuclt get inner class from method by using libCST and ast module

lilypad/_utils/closure.py
```diff
@@ -795,8 +849,45 @@ def from_fn(cls, fn: Callable[..., Any]) -> Closure:
...
+        **kwargs: Any,
+    ) -> Any:
+        """Run the closure instance method.
+
+        Args:
+            *args: Positional arguments to pass to the closure.
+            _init_args: Positional arguments to pass to the closure class constructor.
+            _init_kwargs: Keyword arguments to pass to the closure class constructor.
+            **kwargs: Keyword arguments to pass to the closure.
+        """
```
koxudaxi, line 866: Add new method to run insatnce_method

willbakst: i'm thinking we might want to actually just remove the run methods here and only allow running managed generations (where code needs to be run by the user).

we can always add this later

willbakst (approved):

lilypad/_utils/closure.py
koxudaxi: So, this means we only need to manage prompts and the presence of tools in 'managed', and we don't need to run the actual code in a separate script?

That definitely makes sense, so I'll proceed with removing it.

willbakst (commented):
lgtm, resolve conflicts and i'll approve

koxudaxi:
@willbakst 
I fixed small issue after I merged main.

willbakst (approved):

Issue #258, opened 2025-02-27
Access information for a specific generation function call


willbakst:
we should implement a `Generation` class that contains all of the information for a generation and include an argument that enables returning the `Generation` class instead of the original response. This would make it possible to do things down the line like forward annotations for a specific generation / trace to the user's user. For example, they may want to call the generation, handle the response, and then also let their user annotate the trace, which is not currently possible.

I'm thinking we add an argument e.g. `mode: GenerationMode = "no-wrap"` with an option "wrap" that forces the `Generation` return type. We would then make it generic on the original type (i.e. `Generation[Message]`) where the user can access the message / original response through something like a `output` field or something.

WDYT? I'm not set on the naming, so ideas are welcome here

_Originally posted by @willbakst in https://github.com/Mirascope/lilypad/pull/207#discussion_r1971935105_

willbakst:
This is implemented and will be released along with managed generations

PR #259, opened 2025-02-27
feat: add generation wrap mode

koxudaxi:
Closes: https://github.com/Mirascope/lilypad/issues/258


lilypad/generations.py
```diff
@@ -38,6 +49,37 @@
...
+    WRAP = "wrap"
+
+
+T_co = TypeVar("T_co", covariant=True)
+
+
+class Generation(Generic[T]):
+    """Container for a generation output and its metadata."""
+
+    def __init__(self, output: T, metadata: GenerationPublic) -> None:
```
willbakst, line 68: we also need to include the information for the specific trace that was run. this only gives the generation metadata, but we'll want access to the trace/span id so that we can do things with that specific run (and not just the generation metadata)

lilypad/generations.py
```diff
@@ -386,8 +400,10 @@ def inner(*args: _P.args, **kwargs: _P.kwargs) -> _R:
                     output = fn(*args, **kwargs)
                     if isinstance(output, BaseModel):
                         output = str(output.model_dump())
+                    # Capture trace/span ID
+                    trace_id = format(span.get_span_context().trace_id, "x")
```
koxudaxi, line 404: @willbakst 
Is this what you mean by trace_id in your comment?
However, it may not be possible to get it at this pre-commit stage, so you may need to delay and get it in the Genration class.
And as for the span_id, are you talking about the DB or the ID of the span here?

koxudaxi: I checked at hand and found that trace_id and span_id could be obtained in this way, so add span_id as well.

lilypad/generations.py
```diff
@@ -276,7 +414,11 @@ def inner(*args: _P.args, **kwargs: _P.kwargs) -> _R:
                     if isinstance(output, BaseModel):
                         output = str(output.model_dump())
                     span.set_attribute("lilypad.generation.output", str(output))
-                return output  # pyright: ignore [reportReturnType]
```
koxudaxi, line 279: @willbakst @brenkao 
I guess this is a bug.
I think we should not change the retrun value to str

willbakst: i thought you fixed this in #246?

willbakst: must be that the `managed-generations` branch is out of sync with main and you branched from this branch?

koxudaxi: Oh I think so that I fixed.
We should update managed-generations branch

willbakst: we should be mindful when working on feature branches to sync with main before creating any new branches so we ensure the feature branch stays up to date (and reduce conflicts)

koxudaxi: OK, I merge the main to managed-generations branch.
I tested it in local and run unittest , formatter and lint.
but, we should check again before merge. because, many conflict happened when merged.

willbakst (approved):

lilypad/generations.py
willbakst: ok, i'll merge this into managed generations, and we can do a more formal / strict check when we're merging into main

PR #263, opened 2025-02-28
feat: generate annotations with dummy data

brenkao:


lilypad/ee/server/generate.py
```diff
@@ -0,0 +1,59 @@
...
+)
+@prompt_template(
+    """
+    Please generate a JSON output that complies with the following JSON Schema. 
+    The output should only include valid JSON data according to the schema and 
+    should not contain any additional text or markdown formatting.
+
+    Please provide an example output that fits this schema.
+
+    - idealOutput: str
```
willbakst, line 40: this is hilarious, if the judge outputs a good idealOutput it should just be the original call lol

we should think through a flow here where the judge only outputs reasoning / label

if exact is true, then we don't need the judge. if there's no ideal output, then we should probably just pull the current output into ideal output if the label is pass. if there's an existing ideal output, we should overwrite it in the event that we detect that the current answer is better.

willbakst: this is mostly a note-to-self as i know this is dummy for POC

lilypad/server/api/v0/spans_api.py
```diff
@@ -67,4 +70,55 @@ async def get_aggregates_by_project_uuid(
     return span_service.get_aggregated_metrics(project_uuid, time_frame=time_frame)
 
 
+@spans_router.get("/projects/{project_uuid}/spans/{span_uuid}/generate")
+async def stream_generation(
```
willbakst, line 74: i'm a little confused why this is living in spans? wouldn't this live in annotations api?

lilypad/ee/server/generate.py
willbakst, line 1: nit: let's structure this is `lilypad/ee/server/generations/annotate_trace.py`

lilypad/ee/server/generate.py
```diff
@@ -0,0 +1,59 @@
...
+                    "idealOutput": "value",
+                    "reasoning": "value",
+                    "exact": True,
+                    "label": "pass"
+                }}
+            }}
+        }}
+    """
+)
+async def dummy_prompt() -> None:
```
willbakst, line 58: nit: in line with structure comment, let's name this `annotate_trace` even though it's dummy

lilypad/ee/server/generate.py
brenkao: yeah this is mostly POC so it can stream idealOutput, etc.

lilypad/server/api/v0/spans_api.py
brenkao: its a little weird because when you're creating an annotation, you're doing so from a trace. so we don't have any annotation yet, and we need the span information. but if you're updating a trace, you're doing so from annotation. since streaming is only for get requests (for now), i opted to do it this way, since we can't post data.

willbakst: ideally we would do something like `/projects/{project_uuid}/spans/{span_uuid}/annotate_trace` and have this endpoint live inside of `annotations_api` which is `ee` (and this is not)

willbakst: in this vein this route should have the license checker

lilypad/ee/server/api/v0/annotations_api.py
```diff
@@ -116,3 +119,55 @@ async def get_annotations(
             generation_uuid
         )
     ]
+
+
+@annotations_router.get("/projects/{project_uuid}/spans/{span_uuid}/generate")
```
willbakst, line 124: let's make the route more specific and the function name more specific

for example, `"/projects/{project_uuid}/spans/{span_uuid}/generate_annotation"`

and then the function below should be `generate_annotation`

willbakst (approved):

PR #266, opened 2025-03-03
add: Add opentelemetry-google-genai

koxudaxi:


brenkao:
After telemetry, I have this line
https://github.com/Mirascope/lilypad/blob/88e86086c504210a3abdca803bad265ba697e857/lilypad/server/_utils/spans.py#L61
If the gemini args and response are different from the google package, can you update/add a `convert_google_message` fn

koxudaxi:
@brenkao 
OK, I have updated the attributes for the function.

brenkao (approved):

PR #267, opened 2025-03-03
add: add trace-decorator

koxudaxi:


lilypad/server/client.py
```diff
@@ -257,6 +257,42 @@ def get_or_create_generation_version(
                 },
             )
 
+    def get_or_create_trace_generation(
```
brenkao, line 260: What is this for? I don't think we are saving `lilypad.traces`.

lilypad/_utils/fn_is_async.py
brenkao, line 1: nit: maybe rename this to be like `common` or `helpers` so its more general

willbakst: hmm, for now why don't we just convert `_utils` into `_utils.py` and put all of these utilities in that single file?

we can always refactor later if the utilities file becomes bloated and needs refactoring into sub-modules

lilypad/generations.py
willbakst, line 1: how are we going to handle these changes with the `managed-generations` branch?

i'm wondering if we should implement this as part of that feature branch?

lilypad/prompts.py
willbakst, line 1: same there here, since we're removing prompts, this seems unnecessary (if we were basing on `managed-generations` this wouldn't exist)

lilypad/server/client.py
willbakst: yeah, I think the `SpanTable` should already have everything we would need since we're not doing any code versioning.

Although it's worth asking -- is there a world in which we want to create a `TraceTable` or something with the code? For example, maybe I want to refactor my prompt creation into a method and use `lilypad.trace` for structure and versioning but not treat it as a generation (because it isn't one)

brenkao: I would update the Generation table to be even more generic then, and add a type to it, e.g. type = 'generation'. type = 'function'

lilypad/generations.py
brenkao: that would be good, most of the stuff i'm touching will be on the client, so there should be little merge conflicts

lilypad/server/client.py
willbakst: so migrate `GenerationTable` -> `TraceTable` with the `type` field?

lilypad/_utils/fn_is_async.py
koxudaxi: > hmm, for now why don't we just convert _utils into _utils.py and put all of these utilities in that single file?

The reason for the new file is to avoid circular references.
It would probably be necessary to delay many imports to combine them into one file.

lilypad/prompts.py
koxudaxi: It's a modification to the batch change.
I have reverted. (the `managed-generations` branch is responsible for deleting files.)

lilypad/generations.py
koxudaxi: > i'm wondering if we should implement this as part of that feature branch?

I thought the `managed-generations` branch would be merged soon, but will it take more time?
If it won't be merged for a while, I'll create a PR from the `managed-generations` branch.

lilypad/_utils/fn_is_async.py
koxudaxi: I deleted this as it was completely forgotten to be removed, but I was doing a thought experiment about whether it might be beneficial to keep something in the database when tracing in the future.

lilypad/server/client.py
koxudaxi: I deleted this as it was completely forgotten to be removed, but I was doing a thought experiment about whether it might be beneficial to keep something in the database when tracing in the future.

lilypad/generations.py
brenkao: it is now merged

brenkao:
After spans are merged in to this feature branch, is this branch ready to go?

willbakst (approved):

PR #269, opened 2025-03-04
feat: Add spans

koxudaxi:


lilypad/server/schemas/spans.py
```diff
@@ -59,14 +59,22 @@ def _convert_span_table_to_public(
     ) -> dict[str, Any]:
         """Set the display name based on the scope."""
         data = span.data
-        attributes = data.get("attributes", {})
+        attributes: dict[str, Any] = data.get("attributes", {})
         if span.scope == Scope.LILYPAD:
-            attributes: dict[str, Any] = span.data.get("attributes", {})
             span_type: str = attributes.get("lilypad.type", "unknown")
-            display_name = span.data.get("name", "")
+            display_name = data.get("name", f"Lilypad Span ({span_type})")
```
brenkao, line 65: why did this display name change?

koxudaxi: Because if the display name is not changed, when using this new span, the message `unknown with unknown` will always appear.

brenkao: I thought the user sets the display name when they do
```
with lilypad.span("my display name") as span:
    ...
```

koxudaxi: @brenkao 
OK, I have fixed it.

Issue #271, opened 2025-03-04
Enable running any generation version by simply specifying the version to run


willbakst:
### Description

It would be awesome to be able to do something like this:

```python
import lilypad
from openai import OpenAI
 
client = OpenAI()
 
# @lilypad.generation()
# def answer_question(question: str) -> str:
#    completion = client.chat.completions.create(
#         model="gpt-4o-mini",
#         messages=[{"role": "user", "content": question}],
#     )
#     return str(completion.choices[0].message.content)
 
@lilypad.generation()
def answer_question(question: str) -> str:
    completion = client.chat.completions.create(
        model="gpt-4o-mini",
        messages=[{"role": "user", "content": f"Answer this question: {question}"}],
    )
    return str(completion.choices[0].message.content)

# Runs the commented out v1 above that we automatically versioned
response = answer_question.version("1")("What is the capital of france?")
```

This should also work with managed generations:

```python
import lilypad

@lilypad.generation(managed=True)
def answer_question(question: str) -> str: ...

# Runs the commented out v1 above that we automatically versioned
response = answer_question.version("1")("What is the capital of france?")
```

In this example, we should be able to run even non-managed generation versions.

koxudaxi:
I just noticed that in the non-managed example, it doesn't use mirascope call, so it won't be possible to restore the LLM call from the DB generations record.
Is this discussion about when wrapping mirascope calls?
Currently, we can restore calls from prompt_template, but it should not be possible to restore LLM calls from the logic inside functions (for security reasons).

willbakst:
For non-managed generations, we can pull the runnable closure and run that (ignoring any existing code). The primary issue to resolve is how to properly match signatures in case the version the user is trying to run requires different arguments compared to the current signature in the code (which is why I mentioned doing some form of signature pulling/syncing/matching)

willbakst:
Essentially. In theory we can just `exec` the closure using a `uv` script (since we have the deps). This was the previous `run` method we had implemented that we removed. Of course, we'll want to likely add some form of safety here.

We could use E2B for a sandbox here as well like we mention in #272

koxudaxi:
I looked into E2B but found that it requires signing up for a service, so I don't think it's very good for users to run in their own environments.

willbakst:
Even just the `Sandbox` code interpreter? Did you follow the link in the other issue?

koxudaxi:
[...]

PR #273, opened 2025-03-05
feat: Enable version-specific generation execution with auto-generated type stubs

koxudaxi:
Closes: https://github.com/Mirascope/lilypad/issues/271

### Basic Usage

```bash
$ uv run lilypad stubs llm-project

Found 2 function(s) with decorator lilypad.generation

Generated stub files for 1 source file(s).
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┓
┃ Source File                        ┃ Functions                         ┃ Versions ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━┩
│ /Users/koudai/llm-project/my_prom… │ answer_question                   │ 6        │
│ /Users/koudai/llm-project/my_prom… │ answer_question,                  │ 2        │
│                                    │ async_answer_question             │          │
```

### Before: Type checking errors

With Pyright:
```bash
$ uv run pyright main.py
/Users/koudai/llm-project/main.py
  /Users/koudai/llm-project/main.py:4:77 - error: Expected 1 positional argument (reportCallIssue)
  /Users/koudai/llm-project/main.py:7:78 - error: Argument of type "Literal['100']" cannot be assigned to parameter "max_lenght" of type "int" in function "__call__"
    "Literal['100']" is not assignable to "int" (reportArgumentType)
2 errors, 0 warnings, 0 informations 
```

With MyPy:
```bash
$ uv run mypy main.py
main.py:4: error: Too many arguments for "__call__" of "AnswerQuestionVersion2"  [call-arg]
Found 1 error in 1 file (checked 1 source file)
```

## Project Structure

```
llm-project/
├── my_prompt/
│   ├── __init__.py
│   ├── calls.py
│   └── calls.pyi  <-- Generated by lilypad stubs
└── main.py
```

### Source File: my_prompt/calls.py

```python
from openai import OpenAI
import lilypad

client = OpenAI()

@lilypad.generation()
def answer_question(question: str) -> str:
    completion = client.chat.completions.create(
        model="gpt-4o-mini",
        messages=[{"role": "user", "content": f"Answer this question: {question} and provide the answer"}],
    )
    return str(completion.choices[0].message.content)

@lilypad.generation()
async def async_answer_question(question: str) -> str:
    completion = client.chat.completions.create(
        model="gpt-4o-mini",
        messages=[{"role": "user", "content": f"Answer this question: {question} and provide the answer"}],
    )
    return str(completion.choices[0].message.content)
```

### Test Implementation: main.py

```python
from my_prompt.calls import async_answer_question, answer_question

answer_question.version(3)("What is the capital of france?", "en")
answer_question.version(forced_version=2)("What is the capital of france?", "invalid args")

async def run_async():
    await async_answer_question.version(2)("What is the capital of france?", "100")
```

### Generated Stub File: my_prompt/calls.pyi

```python
# This file was auto-generated by lilypad stubs command
from typing import Any, Callable, Literal, Protocol, overload


class AnswerQuestionVersion1(Protocol):
    def __call__(self, question: str) -> str: ...


class AnswerQuestionVersion2(Protocol):
    def __call__(self, question: str) -> str: ...


class AnswerQuestionVersion3(Protocol):
    def __call__(self, question: str, language: str) -> str: ...


class AnswerQuestionVersion4(Protocol):
    def __call__(self, question: str) -> str: ...


class AnswerQuestionVersion5(Protocol):
    def __call__(self, question: str, language: str) -> str: ...


class AnswerQuestionVersion6(Protocol):
    def __call__(self, question: str, language: str, max_lenght: int) -> str: ...


class AnswerQuestion(Protocol):
    def __call__(self) -> str: ...

    @classmethod
    @overload
    def version(cls, forced_version: Literal[1]) -> AnswerQuestionVersion1: ...

    @classmethod
    @overload
    def version(cls, forced_version: Literal[2]) -> AnswerQuestionVersion2: ...

    @classmethod
    @overload
    def version(cls, forced_version: Literal[3]) -> AnswerQuestionVersion3: ...

    @classmethod
    @overload
    def version(cls, forced_version: Literal[4]) -> AnswerQuestionVersion4: ...

    @classmethod
    @overload
    def version(cls, forced_version: Literal[5]) -> AnswerQuestionVersion5: ...

    @classmethod
    @overload
    def version(cls, forced_version: Literal[6]) -> AnswerQuestionVersion6: ...

    @classmethod  # type: ignore[misc]
    def version(cls, forced_version: int) -> Callable[..., Any]: ...


answer_question = AnswerQuestion

# This file was auto-generated by lilypad stubs command
from typing import Any, Callable, Coroutine, Literal, Protocol, overload


class AsyncAnswerQuestionVersion1(Protocol):
    def __call__(self, question: str) -> Coroutine[Any, Any, str]: ...


class AsyncAnswerQuestionVersion2(Protocol):
    def __call__(
        self, question: str, max_lenght: int = 100
    ) -> Coroutine[Any, Any, str]: ...


class AsyncAnswerQuestion(Protocol):
    def __call__(self) -> Coroutine[Any, Any, str]: ...

    @classmethod
    @overload
    def version(cls, forced_version: Literal[1]) -> AsyncAnswerQuestionVersion1: ...

    @classmethod
    @overload
    def version(cls, forced_version: Literal[2]) -> AsyncAnswerQuestionVersion2: ...

    @classmethod  # type: ignore[misc]
    def version(
        cls, forced_version: int
    ) -> Callable[..., Coroutine[Any, Any, Any]]: ...


async_answer_question = AsyncAnswerQuestion
```


willbakst:
This is awesome. It might be time for us to start telling users to put everything inside a `generations/` folder

lilypad/_utils/__init__.py
```diff
@@ -6,13 +6,15 @@
 from .fn_is_async import fn_is_async
 from .functions import ArgTypes, ArgValues, create_mirascope_call, inspect_arguments
 from .middleware import create_mirascope_middleware, encode_gemini_part
+from .sandbox import DockerSandboxRunner
```
willbakst, line 9: don't we also need to suppress this in case the user doesn't have docker?

lilypad/generations.py
```diff
@@ -38,6 +40,7 @@
     load_config,
 )
 from ._utils.middleware import SpanContextHolder
+from ._utils.sandbox import DockerSandboxRunner
```
willbakst, line 43: this looks like now users will always have to have `docker` installed so things can run in the sandbox?

given that it's the user who generated the code that's running it, I don't think we actually need the docker sandbox for running generations unless desired?

can we update this so that you can specify (maybe as part of the `version` method) that you want to run it in a sandbox and what sandbox to use? I expect it would take the sandbox runner or something and then the consumer would import the docker sandbox and provide it?

willbakst: I also wonder if this actually needs a two pronged solution (since a user might feel weird about running code pulled from our API)

1. We use a Sandbox to run user-generated code when it's run in our system (which we haven't built yet, this is in prep for that)
2. Instead of pulling generation code on the fly we actually sync the code in their system (like the stubs) as individual files or something? In fact, why even bother with "stubs" when we could just have the actual generation version code stored in some structured format? that way when a user forces a specific version they can feel safe about what's running because they can see the code before they run it / deploy it

lilypad/_utils/__init__.py
koxudaxi: We have a restored implementation of SubprocessSandboxRunner which is being used as the standard.
I'm working on the fixes

lilypad/generations.py
koxudaxi: I think this needs careful consideration.
While stubs only need the interface, when generating implementations, issues with scope and name resolution arise.
For example, in the case mentioned, multiple generations exist in the same file. In this case, we would need to merge all generations into one file.
However, the problem here is that code that depends on generations is also stored in the DB. I think it would be extremely difficult to cleanly merge these (code outside of functions), so we would need to design a post-restoration structure such as separating them into different files.
There are probably other issues that would arise when restoring implementations, but I can't think of any more at the moment. However, I think it's not a bad idea to realize this in the future through gradual verification.

willbakst: Yeah I guess matching the hash would be quite difficult, although now this has me thinking how we're matching the hash with the `.version` method now? At a certain point can't we just pull the `GenerationPublic` and run against that or something?

Food for thought. In the short term the stubs + sandbox is enough I think and we can work to improve things iteratively. The main thing is that the sandbox should not be required unless using the `.version` decorator since users not using `.version` shouldn't have to have e.g. `docker` installed.

koxudaxi:
I discovered that kwargs with default values aren't working when testing various cases. I'm currently fixing this since arg_values also need to be stored in the DB.
Of course, not all types can be stored, but this should be an acceptable constraint for default arguments in generations.

lilypad/generations.py
koxudaxi: Unfortunately, this only matches based on generation.name.
Since everything except the name can be changed, there's no other way to perform the check.
This is my main concern.

koxudaxi: The concern is not about that, but whether users might use the same function name within the same project. For example, if there's a function answer(question) in /parent_module/module_a, there's a worry that users might create another function with the same name answer(question) in /parent_module/module_b. As a project grows larger, it becomes difficult to keep track of which function names are being used across the entire project. Of course, I understand that this won't be an immediate issue since not many people are using generation yet.

willbakst: I don't understand the issue if we're just running the exact code we pulled from the system in a fresh `uv` environment that only has the required deps installed? How would that every clash with anything else? We're just running a stand-alone script, right?

koxudaxi: The issue is not about conflicts during execution, but about users accidentally using the same function names.

koxudaxi:
@willbakst @brenkao 
ok,it's ready for review.

lilypad/generations.py
willbakst: does that matter though?

even if a user had two functions with the same names both using `generation` decorator in separate modules, when they run wouldn't they just get attached to their version?

I still don't understand the issue here

ee/validate.py
```diff
@@ -183,3 +195,84 @@ def generate_license(
...
+    if not cached_license:
+        config = load_config()
+        from lilypad.server.client import LilypadClient
+
+        lilypad_client = LilypadClient(
+            token=config.get("token", None),
+        )
+        cached_license = lilypad_client.get_license_info()
+    if cached_license.tier not in tiers:
+        raise LicenseError(INVALID_LICENSE_MESSAGE)
```
willbakst, line 220: we should make this `LilypadLicenseError` so `call_safely` handles it properly?

also should we just push `INVALID_LICENSE_MESSAGE` as the default message in the error class?

ee/validate.py
```diff
@@ -183,3 +195,84 @@ def generate_license(
...
+    return cached_license
+
+
+@overload
+def require_license(
+    tiers: set[Tier],
+) -> Callable[[Callable[_P, _R]], Callable[_P, _R]]: ...
+@overload
```
willbakst, line 228: nit: add additional space between overloads and main function etc. for better readability

lilypad/ee/generations.py
```diff
@@ -0,0 +1,110 @@
...
+if TYPE_CHECKING:
+    from lilypad.generations import Generation
+
+_MANGED_PROMPT_TEMPLATE: TypeAlias = bool
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+_SandBoxRunnerT = TypeVar("_SandBoxRunnerT", bound=SandboxRunner)
+
+
+def _specific_generation_version_sync_factory(
```
willbakst, line 22: why are these private? generally I expect private methods only ever to be used inside their file.

for methods we want to be private but use internally, we should put them inside of a private module and never export them beyond that. For example, if we want these to private, we should rename the parent file to `_generations.py` but make the methods "public" in that they don't have the underscore

lilypad/generations.py
koxudaxi: What about in the case of managed? Wouldn't one valid version be applied to multiple decorators (functions)? In other words, I feel like a prompt different from the intended one might be executed. Are the functions distinguished from each other on the UI side?
I might be overthinking this a bit.

ee/validate.py
```diff
@@ -183,3 +195,84 @@ def generate_license(
     data_b64 = base64.urlsafe_b64encode(data_bytes).rstrip(b"=").decode("ascii")
     sig_b64 = base64.urlsafe_b64encode(signature).rstrip(b"=").decode("ascii")
     return f"{data_b64}.{sig_b64}"
+
+
+INVALID_LICENSE_MESSAGE = "Invalid License. Contact support@mirascope.com to get one."
+
+
+def _validate_license_with_client(
+    cached_license: LicenseInfo | None, tiers: set[Tier]
```
willbakst, line 204: hmm, why is `tiers` now a `set`?

the tiers are sequential, so can't we just always pass the requirement if the given license is greater than or equal to the required license? am i missing something?

we could make this easy if we update `Tier` to be `Tier(int, Enum)` so we can easily compare the enum values using `>=`

lilypad/ee/server/require_license.py
```diff
@@ -65,7 +65,7 @@ async def wrapper(*args: Any, **kwargs: Any) -> Any:
 class RequireLicense:
     """License dependency for FastAPI endpoints."""
 
-    def __init__(self, tier: Tier = Tier.ENTERPRISE) -> None:
+    def __init__(self, tier: Tier | None = Tier.ENTERPRISE) -> None:
         """Initialize with required tier."""
         self.tier = tier
```
willbakst, line 71: noticing below that we only throw an "Invalid License" error for enterprise mismatch but shouldn't we handle the new cases as well?

also noticing we're using `HTTPException` instead of the Lilypad-specific exception classes?

lilypad/server/client.py
```diff
@@ -123,15 +125,15 @@ def _request(
         endpoint: str,
         response_model: None,
         **kwargs: Any,
-    ) -> dict[str, Any]: ...
+    ) -> dict[str, Any] | Any: ...
```
willbakst, line 128: why add `Any` here?

lilypad/generations.py
```diff
@@ -283,30 +372,28 @@ def _construct_trace_attributes(
...
-    generation_ = get_generation(arg_types)
+    generation_: GenerationPublic,
+) -> Generator[None, None, None]:
     token = current_generation.set(generation_)
     try:
         # Check if this is the outermost generation (no previous generation)
         is_outermost = token.old_value == Token.MISSING
         with _outermost_lock_context(is_outermost):
-            yield arg_types, arg_values, generation_
+            yield
```
willbakst, line 396: why do we no longer need to yield anything here? don't we need `GenerationPublic` when we want the generation info through wrap?

lilypad/generations.py
```diff
@@ -370,6 +457,38 @@ def sync_wrapper() -> Message | Stream:
...
+        sandbox_factory = SubprocessSandboxFactory()
+    closure = Closure(
+        name=generation.name,
+        code=generation.code,
+        signature=generation.signature,
+        hash=generation.hash,
+        dependencies=generation.dependencies,
+    )
+
+    sandbox = sandbox_factory.create(closure)
```
willbakst, line 476: i feel like it would be better for this method to accept a sandbox instead of a factory so that a user could provide a reusable sandbox if they want rather than having to create a new sandbox every time? and then `execute_function` would take the function along with `*args, **kwargs`?

lilypad/generations.py
```diff
@@ -456,7 +580,11 @@ def decorator(
         if fn_is_async(fn):
 
             def _create_inner_async(
-                get_generation: Callable[[_ArgTypes], GenerationPublic],
+                get_generation: Callable[
+                    _P,
+                    tuple[GenerationPublic, _MANGED_PROMPT_TEMPLATE],
+                ],
+                inner_sandbox_factory: SandBoxFactory | None,
```
willbakst, line 587: can you explain why the standard generation needs a sandbox? wouldn't only the version method need it?

lilypad/generations.py
```diff
@@ -515,73 +642,63 @@ async def _inner_async(
...
-                        f"Generation version {forced_version} not found for function: {fn.__name__}"
-                    )
-
-                def _get_specific_version(_: _ArgTypes) -> GenerationPublic:
-                    return specific_version_generation
-
-                return _create_inner_async(_get_specific_version)
-
-            inner_async.version = version_async  # pyright: ignore [reportAttributeAccessIssue, reportFunctionMemberAccess]
+            inner_async.version = _specific_generation_version_async_factory(  # pyright: ignore [reportAttributeAccessIssue, reportFunctionMemberAccess]
```
willbakst, line 669: same thing here, i feel like the only time a user should ever have to think about a sandbox is when they call the `.version` method, which would accept a sandbox within which the returned version would run?

lilypad/server/client.py
koxudaxi: I forget to delete it.

lilypad/generations.py
koxudaxi: Because of the change to pass generation_ as an external argument instead of getting it inside the contextmanager.

lilypad/generations.py
koxudaxi: The reason for creating factroy is to correctly pass arguments specific to each sandbox.
Currently, if it is docker, image. env if it is a shared one.
Of course, you can pass a sandbox instance, but sandbox was assumed to be a one-time clean environment.
If I pass a sandobx instance, should it be sandbox_instance.execute_function(_closure=closure, *args, **kwargs)?
One of the reasons why I prepared factory is that the argument of kwargs could be covered by closure.

lilypad/generations.py
koxudaxi: If `manged=True`, shouldn't it be done in the sandbox as with version?
Rather, we need to consider how it should be executed the code from string in DB if there is no sandbox.

willbakst: no if `managed=True` then we can just run the mirascope call under the hood, which is safe. It's only when we're executing potentially arbitrary code, which happens during `.version` when we pull a non-managed version. Right?

lilypad/generations.py
willbakst: I see, if it doesn't add a ton of additional overhead, I can understand using the factory. I figured that if the user does not provide a sandbox we would likely still create a new one every time and pass that in. It would only be structured as sandbox instead of sandbox factory so that a user could maintain a sandbox to reduce overhead / improve performance.

lilypad/generations.py
koxudaxi: I see, I misunderstood a bit.
So it's only in the case of .version where execution without using mirascope is possible.
I'll fix it.

lilypad/generations.py
koxudaxi: OK, same above

lilypad/server/client.py
koxudaxi: fixed

lilypad/ee/server/require_license.py
koxudaxi: > noticing below that we only throw an "Invalid License" error for enterprise mismatch but shouldn't we handle the new cases as well?

OK, I fixed it .

> also noticing we're using HTTPException instead of the Lilypad-specific exception classes?

I expect the Lilypad-specific exception classes are for only client side.

koxudaxi: I added the exceptions.py for server

lilypad/generations.py
koxudaxi: I deleted sandbox factory

lilypad/generations.py
koxudaxi: I have fixed it.

lilypad/generations.py
koxudaxi: I have fixed it.

ee/validate.py
koxudaxi: I changed the str, enum to int ,enum

lilypad/ee/server/require_license.py
koxudaxi: I fixed it.

lilypad/ee/generations.py
koxudaxi: I fixed it.

ee/validate.py
koxudaxi: I created LilypadLicenseError

lilypad/cli/commands/stubs.py
```diff
@@ -0,0 +1,460 @@
...
+    header_types = (
+        "overload, Literal, Callable, Any, Protocol, Coroutine"
+        if is_async
+        else "overload, Literal, Callable, Any, Protocol"
+    )
+    joined_protocols = "\n\n".join(version_protocols)
+    joined_overloads = "\n\n".join(version_overloads)
+    content = f"""# This file was auto-generated by lilypad stubs command
+from typing import {header_types}
+from lilypad._utils.sandbox import SandboxRunner
```
koxudaxi, line 303: Is this a reasonable location?

lilypad/generations.py
willbakst: I think you're overthinking this. Let's sync when you're up to resolve

ee/validate.py
willbakst: I may have been wrong here. Discussing with @brenkao about this.

We really only need the Lilypad-specific errors for when users are running code that could have a Lilypad failure but should still be able to run (i.e. `trace` / `generation` decorators), but for things that should fail (like accessing an enterprise protected feature) then we don't need that safety (cause it should fail)

lilypad/generations.py
koxudaxi: You're right. I was overthinking it. Let's consider this resolved.

lilypad/cli/commands/stubs.py
koxudaxi: I was checking if this is located in an appropriate module. I see it's importing `from lilypad._utils.sandbox` (with an underscore), but since users are accessing it directly, I was wondering if we should move it to a different module without the underscore prefix

lilypad/_utils/sandbox/runner.py
```diff
@@ -11,35 +11,36 @@ class SandboxRunner(ABC):
...
-    def __init__(
-        self, closure: Closure, environment: dict[str, str] | None = None
-    ) -> None:
-        self.closure: Closure = closure
+    def __init__(self, environment: dict[str, str] | None = None) -> None:
         self.environment: dict[str, str] = environment or {}
 
     @abstractmethod
-    def execute_function(self, *args: Any, **kwargs: Any) -> str: ...
+    def execute_function(self, _closure: Closure, *args: Any, **kwargs: Any) -> str: ...
```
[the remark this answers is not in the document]
koxudaxi: I renamed it to `_closure` to avoid interference with any closure parameter that might exist in the generation arguments. Do you think we should name it closure instead?

lilypad/_utils/sandbox/runner.py
```diff
@@ -11,35 +11,36 @@ class SandboxRunner(ABC):
...
+    @classmethod
+    def _is_async_func(cls, closure: Closure) -> bool:
+        lines = closure.signature.splitlines()
         return any(
             line.strip() for line in lines if line.strip().startswith("async def ")
         )
 
-    def _generate_async_run(self, *args: Any, **kwargs: Any) -> str:
+    @classmethod
+    def _generate_async_run(cls, _closure: Closure, *args: Any, **kwargs: Any) -> str:
```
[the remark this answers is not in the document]
koxudaxi: same above

lilypad/generations.py
```diff
@@ -381,7 +380,7 @@ def _construct_trace_attributes(
     }
 
 
-_MANGED_PROMPT_TEMPLATE: TypeAlias = bool
+_ManagedPromptTemplate: TypeAlias = bool
```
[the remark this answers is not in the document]
koxudaxi: OK, you're right.

ee/validate.py
koxudaxi: OK, I will revert the changes.

koxudaxi: I reverted

lilypad/_utils/sandbox/runner.py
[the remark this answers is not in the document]
koxudaxi: I changed it to `closure`

lilypad/_utils/sandbox/runner.py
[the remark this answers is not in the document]
koxudaxi: I changed it to `closure`

lilypad/generations.py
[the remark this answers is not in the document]
koxudaxi: I fixed it

lilypad/cli/commands/stubs.py
willbakst: ah yes I see, I wasn't thinking of this as user code but you're right it is.

Maybe we should move it inside something like `lilypad.sandbox`?

koxudaxi: OK, I have moved it to `lilypad.sandbox`

koxudaxi:
@willbakst 
It's ready for review.

willbakst (approved):

PR #290, opened 2025-03-07
Managed generation traces

### LINKED ISSUE #285: Managed Generation shows the wrong code
### Description

For an `answer_question` generation with a `question: str` input, I'm getting this code:

```python
import lilypad


@lilypad.prompt()
def answer_question(question: str): ...
```

This is incorrect and should instead be the following:

```python
import lilypad


@lilypad.generation(managed=True)
def answer_question(question: str) -> lilypad.Message: ...
```

and this much perfectly match the computed closure when the user runs the code locally (so we properly version match)

### Python, Lillypad & OS Versions, related packages (not required)

```TOML
python-lilypad==0.0.30
```

### LINKED ISSUE #286: Traces for managed generations when run in the UI
### Description

Right now running a managed generation in the UI does not have a corresponding trace in the traces tab, but this goes against the principle of always tracing every LLM API call.

We should make sure that any time a generation is run (whether it's in code or in the UI) the run is properly versioned and traced.

brenkao:
Closes #286 
Closes #285


lilypad/ee/server/api/v0/generations_api.py
```diff
@@ -0,0 +1,215 @@
...
+        generation_create.arg_types or {}, generation_create.name, False
+    )
+    generation_create.version_num = generation_service.get_next_version(
+        project_uuid, generation_create.name
+    )
+    generations = generation_service.find_generations_by_signature(
+        project_uuid, generation_create.signature
+    )
+    if len(generations) == 0:
+        generation_create.is_default = True
```
koxudaxi, line 73: Is it appropriate to use is_default if not deployed in a specific environment?
I don't want to add more work, but I don't want to use too implicit behavior.

brenkao: this is placeholder for once the deployment PR is implemented. I figured it wasn't worth making a version that works when soon it will need to be updated.

koxudaxi (approved):

lilypad/ee/server/api/v0/generations_api.py
```diff
@@ -0,0 +1,215 @@
...
+    api_key_service: Annotated[APIKeyService, Depends(APIKeyService)],
+) -> str:
+    """Run version."""
+    generation = generation_service.find_record_by_uuid(generation_uuid)
+    api_keys = api_key_service.find_keys_by_user_and_project(project_uuid)
+    if len(api_keys) == 0:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST, detail="No API keys found"
+        )
+    generation_version = (
```
willbakst, line 96: nit: naming is weird here, maybe `specific_generation_version_fn`?

PR #303, opened 2025-03-07
fix: annotation tab tier not working

brenkao:


client/src/hooks/use-isEnterprise.tsx
```diff
@@ -2,9 +2,27 @@ import { Tier } from "@/ee/types/types";
...
+
+export const hasFeatureAccess = (
+  userTier: Tier,
+  featureTier: Tier
+): boolean => {
+  if (userTier === Tier.ENTERPRISE) {
+    return true;
+  }
+
+  return tierLevels[userTier] >= tierLevels[featureTier];
```
willbakst, line 27: does this cover the above case?

willbakst (approved):

Issue #281, opened 2025-03-06
Remove unnecessary dependencies (that only the server requires) in the production package


willbakst:
### Description

We should only include packages necessary for the SDK (not the server) in the production release as required so users using the cloud hosted version don't have the server dependencies installed

willbakst:
Also seeing this:

```
DependencyConflict: requested: "aioboto3>=13.2.0,<14" but found: "aioboto3 14.1.0"
```

willbakst:
We need to make sure that our workflows and other e.g. deployment scripts properly install all dependencies using the additional group for server deps

PR #308, opened 2025-03-10
feat: Add server dependencies

koxudaxi:
Closes: https://github.com/Mirascope/lilypad/issues/281


pyproject.toml
```diff
@@ -33,8 +33,7 @@ classifiers = [
 dependencies = [
     "mirascope>=1.20.1",
     "sqlmodel>=0.0.22",
-    "psycopg2-binary>=2.9.9",
-    "fastapi[standard]>=0.114.0",
+    "fastapi>=0.114.0",
```
koxudaxi, line 36: middleware uses `fastapi` 
https://github.com/Mirascope/lilypad/blob/bd6650ddbfb93acfe9cfa6b1ab515a73c68da368/lilypad/_utils/middleware.py#L13

willbakst: would it be better instead to just implement an encoder rather than require fastapi?

pyproject.toml
```diff
@@ -33,8 +33,7 @@ classifiers = [
 dependencies = [
     "mirascope>=1.20.1",
     "sqlmodel>=0.0.22",
```
willbakst, line 35: do we need sql model if we're not using anything in the server? can't we push this into the server deps?

pyproject.toml
```diff
@@ -46,12 +45,12 @@ dependencies = [
     "ruff>=0.9.5",
```
willbakst, line 45: which of the above crypto packages will a user need if using standard lilypad SDK?

lilypad/server/_utils/__init__.py
```diff
@@ -18,6 +12,15 @@
...
 
+with suppress(ImportError):
+    from .auth import (
+        api_key_header,
+        create_api_key,
+        create_jwt_token,
+        get_current_user,
+        validate_api_key_project_no_strict,
+        validate_api_key_project_strict,
+    )
```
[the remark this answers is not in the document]
willbakst: hmm, i feel like theoretically we shouldn't have to suppress anything inside of the server folder (since we would expect the server extra to be applied)?

i think the main thing is that we should either (1) remove `server` from the main init or (2) suppress it's import so it only throws an error if a user tries to access it?

koxudaxi:
@willbakst 
> would it be better instead to just implement an encoder rather than require fastapi?

Does this mean we implement our own?
Or do we copy the code from fastapi, including the license?

pyproject.toml
koxudaxi: Does this mean we implement our own?
Or do we copy the code from fastapi, including the license?

pyproject.toml
koxudaxi: For example, we use GenerationPublic in our clients and generations, which inherits from SQLModel. In other words, the client side also depends on SQLModel.

pyproject.toml
koxudaxi: It seems unnecessary since we have a new licenseAPI. I will delete it.

willbakst: we should keep whatever packages we need for the validation, move any we still need for auth (if any) into server, and delete any we aren't using

then for the ones we need for validation we should put them in an `ee` extra rather than required?

lilypad/server/_utils/__init__.py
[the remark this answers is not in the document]
koxudaxi: Because of the potato dependency, the error must be suppressed somewhere.
```python
Traceback (most recent call last):
  File "/Users/koudai/Library/Application Support/JetBrains/PyCharm2024.3/scratches/scratch_199.py", line 5, in <module>
    import lilypad
  File "/Users/koudai/PycharmProjects/lilypad/lilypad/__init__.py", line 6, in <module>
    from ._configure import configure
  File "/Users/koudai/PycharmProjects/lilypad/lilypad/_configure.py", line 17, in <module>
    from ._utils import load_config
  File "/Users/koudai/PycharmProjects/lilypad/lilypad/_utils/__init__.py", line 7, in <module>
    from .functions import ArgTypes, ArgValues, create_mirascope_call, inspect_arguments
  File "/Users/koudai/PycharmProjects/lilypad/lilypad/_utils/functions.py", line 26, in <module>
    from ..server.schemas.generations import GenerationCreate, GenerationPublic, Provider
  File "/Users/koudai/PycharmProjects/lilypad/lilypad/server/schemas/__init__.py", line 13, in <module>
    from .spans import SpanCreate, SpanMoreDetails, SpanPublic, SpanTable
  File "/Users/koudai/PycharmProjects/lilypad/lilypad/server/schemas/spans.py", line 14, in <module>
    from .._utils import (
  File "/Users/koudai/PycharmProjects/lilypad/lilypad/server/_utils/__init__.py", line 1, in <module>
    from .auth import (
  File "/Users/koudai/PycharmProjects/lilypad/lilypad/server/_utils/auth.py", line 10, in <module>
    from jose import JWTError, jwt
ModuleNotFoundError: No module named 'jose'
```

pyproject.toml
willbakst: hmm, shouldn't we make `GenerationPublic` just a `BaseModel`? I'm assuming we rely on `SQLModel` so we don't have to pick and choose fields and can inherit them, but wouldn't it make the most sense to always be explicit about every field we make public and remove the dependency on SQLModel in the user facing SDK?

pyproject.toml
willbakst: i think we could copy the code with the license in it's own file and explain we do so to remove the dependency?

WDYT?

willbakst: another option would be to ask claude or something to just generate one that we maintain, but i'm not sure that's worth it given this already exists and we know works with fastapi?

willbakst: the other option would be to take another look at why we're using jsonable_encoder at all and determine if there's a better/easier solution that doesn't need it?

pyproject.toml
koxudaxi: I moved it.

pyproject.toml
koxudaxi: Regardless of fastapi, I think our solution needs a jsonable_encoder.
The reason is that we need to store function arguments and results in a DB or return them via API. So, we could build and maintain our own, even if it costs a little more.

lilypad/server/_utils/__init__.py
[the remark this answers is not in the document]
willbakst: oh i see because we are importing from server in user code

is there a way around this? or should we just suppress what we need to?

pyproject.toml
willbakst: ok, i think it's best to remove the dependency and implement our own, so let's do that

pyproject.toml
```diff
@@ -33,25 +33,22 @@ classifiers = [
...
     "mirascope>=1.20.1",
     "sqlmodel>=0.0.22",
-    "psycopg2-binary>=2.9.9",
-    "fastapi[standard]>=0.114.0",
+    "fastapi>=0.114.0",
     "opentelemetry-api>=1.30.0",
     "opentelemetry-sdk>=1.30.0",
     "opentelemetry-instrumentation>=0.51b0",
     "typer>=0.12.5",
     "itsdangerous>=2.2.0",
```
willbakst, line 41: do we use this anywhere anymore? I can't find and import using it

koxudaxi: OK, I have removed it.

pyproject.toml
willbakst: I think right now an example for what I'm talking about is that `GenerationPublic` and `GenerationCreate` live in the `schema/` directory and inherit from `_GenerationBase`.

I think instead of this we should just write `GenerationPublic` and `GenerationCreate` as pure `BaseModel` subclasses (rather than inheriting from `_GenerationBase`.

This will require duplication of the field definitions, but I think that's the correct approach. I don't think `GenerationPublic` or `GenerationCreate` or anything in `schemas/` for that matter should rely on `SQLModel` -- they should all just be `BaseModel` subclasses.

lilypad/server/_utils/__init__.py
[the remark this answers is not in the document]
koxudaxi: We need to import `_utils.spans`, but we need to import `_utils.auth` at that time, so we need to do import error suppress of auth itself as it like the changes

pyproject.toml
koxudaxi: OK, I will do it.

lilypad/server/_utils/__init__.py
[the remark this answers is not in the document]
koxudaxi: OK, I the `from jose import JWTError, jwt` import statemt in to the  functions for lazy import

pyproject.toml
koxudaxi: OK, I implemented it.

pyproject.toml
koxudaxi: I create public model with BaseModel.

koxudaxi: @willbakst @brenkao 
The existing SQLModel-based Schemas were moved to the Models directory, which contains models that inherit from BaseModel from the Schemas directory.
There are some places where SQLModel is used, such as around user, so please review it.

brenkao: That is just going to make it more difficult for us to maintain. I'm not opposed if you think the tradeoff is worth it.

willbakst: I think it's worth the additional burden. I also think it makes more sense not to assume that the public facing schemas always include everything in base. This means we can add things to base that make sense to be in base but not necessarily the public facing schema

pyproject.toml
willbakst: Hmm, i'm not seeing it? Also not seeing the move of `fastapi` into the `server` extra that would come with this change?

pyproject.toml
koxudaxi: OK, Sounds good. I will do it.

lilypad/server/client/schemas/v0.py
```diff
@@ -0,0 +1,487 @@
...
+    """
+
+    role: UserRole
+    user_uuid: UUID = Field(..., title='User Uuid')
+    uuid: UUID = Field(..., title='Uuid')
+    organization_uuid: UUID = Field(..., title='Organization Uuid')
+    organization: OrganizationPublic
+
+
+class UserOrganizationTable(BaseModel):
```
willbakst, line 419: noticing a few items here that we probably don't need

is it easier to keep them or should we add filters?

pyproject.toml
willbakst: Ok looks like we can move sqlmodel into the server deps now, resolve this once you do that

pyproject.toml
```diff
@@ -69,6 +65,13 @@ server = [
     "resend>=2.6.0",
     "alembic-postgresql-enum>=1.6.0",
     "starlette>=0.26.0",
+    "python-jose[cryptography]>=3.3.0",
+]
+ee = [
+     "cryptography>=43.0.3",
+]
+code-gen = [
```
willbakst, line 73: nit: why not just put this in the dev section?

koxudaxi: I have moved it to dev.

lilypad/server/client/schemas/v0.py
koxudaxi: I will try it

koxudaxi: I did

scripts/generate_client_schema.py
```diff
@@ -0,0 +1,203 @@
...
+    "OrganizationPublic",
+    "ProjectPublic",
+    "SpanPublic",
+    "GenerationCreate",
+    "Provider",
+]
+EE_V0_TARGET_MODELS = []
+
+
+class LilypadOpenAPIParser(OpenAPIParser):
```
koxudaxi, line 42: I implemented a slightly tricky solution to output only specific models. Since this project only deals with simple OpenAPI files, I believe it won't cause any issues.
If time permits, I'll add this feature to the mainstream branch.

ee/validate.py
```diff
@@ -45,9 +46,14 @@ def _missing_(cls, value: Any) -> "Tier":
...
-class LicenseError(Exception):
-    """Custom exception for license-related errors"""
+    @classmethod
+    def __get_pydantic_json_schema__(
+        cls, core_schema: CoreSchema, handler: GetJsonSchemaHandler
+    ) -> JsonSchemaValue:
+        json_schema = handler(core_schema)
+        json_schema = handler.resolve_ref_schema(json_schema)
+        json_schema["x-enum-varnames"] = [f"{choice.name}" for choice in cls]
+        return json_schema
```
koxudaxi, line 56: @brenkao We can generate custom openapi schema by using the method.

client/src/components/HomeSettings.tsx
```diff
@@ -23,10 +23,10 @@ import { SettingsIcon } from "lucide-react";
...
 const tier = {
-  [Tier.VALUE_0]: "Free",
-  [Tier.VALUE_1]: "Pro",
-  [Tier.VALUE_2]: "Team",
-  [Tier.VALUE_3]: "Enterprise",
+  [Tier.FREE]: "Free",
+  [Tier.PRO]: "Pro",
+  [Tier.TEAM]: "Team",
+  [Tier.ENTERPRISE]: "Enterprise",
 };
```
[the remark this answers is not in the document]
brenkao: The issue with this is that int enums get generated as enum: [0,1,2,3] in OpenAPI spec. So my typescript converter translates those to be `VALUE_0` since the `FREE` gets dropped.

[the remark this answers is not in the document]
koxudaxi: I know the issue. But, I set `x-enum-varnames` in the schema.
After that, the typescript generator will generate the names automatically .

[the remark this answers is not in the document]
koxudaxi: https://github.com/acacode/swagger-typescript-api/blob/7fc6c8506ce81c7e1c942290c937cd3e48a2d904/src/schema-parser/schema-utils.ts#L38-L43

pyproject.toml
```diff
@@ -33,25 +33,21 @@ classifiers = [
 dependencies = [
     "mirascope>=1.20.1",
     "sqlmodel>=0.0.22",
```
willbakst, line 35: move this into the `server` extra now that we can

koxudaxi: I have moved sqlmodel to server

PR #309, opened 2025-03-10
feat: Add environment-based deployment system for managed generations

koxudaxi:
# Lilypad Deployment Flow Documentation

## Overview

Lilypad implements a deployment system similar to general cloud configuration service, allowing for safe management of LLM prompts (Generations) across different environments. This documentation outlines the deployment architecture and common workflows.

### Generation
- Contains LLM prompt configurations, parameters, and model specifications
- Versioned with `managed=True` flag for SDK integration
- Stored in database with unique version numbers

### Environment
- Represents deployment targets (production, staging, development)
- Each has a unique API key for authentication
- Always references the currently active deployment

### Deployment
- Maps a specific Generation version to an Environment
- Only one active deployment exists per environment
- Maintains version history for rollback capabilities

## Authentication Methods

1. **Web UI Access**:
   - OAuth2 authentication (GitHub/Google)
   - Full management capabilities for Generations and Deployments
   - Visual interface for deployment operations

2. **SDK Access**:
   - Environment-specific API key authentication
   - Retrieves managed Generations (`managed=True`)
   - Used by applications to access active prompts

## Standard Deployment Flow (Dev to Prod)

1. Create Generation in Web UI (prompt text, model name, parameters)
2. Deploy v1 Generation to development environment
3. Add `@lilypad.generations(managed=True)` decorator to application functions
4. Deploy application to development environment
5. Test thoroughly in development
6. Deploy v1 Generation to production environment
7. Deploy application to production environment

## Manual Rollback Flow

1. Deploy application to development environment
2. Deploy v2 Generation to development environment for testing
3. Detect error traces from application
4. Roll back application if necessary
5. Roll back Generation from v2 to v1 (re-deploy v1)

## Design Principles

- **Explicit over Implicit**: Deployments must be explicitly managed to avoid unexpected behavior
- **Minimalist Start**: Initially using fully manual deployment management
- **Security Focus**: Preventing test prompts from leaking to production environments
- **Future Extensibility**: Design allows for future implementation of advanced deployment strategies

## Future Enhancements

- **Deployment Strategies**: Adding canary releases through `deployment_strategy` field
- **Automatic Rollbacks**: Implementing error detection and conditional rollbacks 
- **Default Behaviors**: Potential fallback to latest Generation during beta if none deployed

## Implementation Notes

- The current design prioritizes explicit control over convenience
- API keys are environment-specific to enforce security boundaries
- Missing deployments result in errors rather than automatic fallbacks
- Future options for auto-deployment may be added but will be opt-in

image

image

CC: @brenkao


lilypad/ee/server/models/deployments.py
```diff
@@ -0,0 +1,39 @@
...
+class DeploymentBase(SQLModel):
+    """Base Deployment Model."""
+
+    environment_uuid: UUID = Field(
+        foreign_key=f"{ENVIRONMENT_TABLE_NAME}.uuid", ondelete="CASCADE"
+    )
+    generation_uuid: UUID = Field(
+        foreign_key=f"{GENERATION_TABLE_NAME}.uuid", ondelete="CASCADE"
+    )
+    is_active: bool = Field(default=True, index=True)
```
brenkao, line 29: i assume this replaces `is_default` inside generations, so we should probably delete that

koxudaxi:
After reviewing it, it looks like I should change the API call when `lilypad.generation(managed=True)`.

lilypad/ee/server/models/environments.py
```diff
@@ -0,0 +1,38 @@
...
+
+if TYPE_CHECKING:
+    from ....server.models.projects import ProjectTable
+    from .deployments import DeploymentTable
+
+
+class EnvironmentBase(SQLModel):
+    """Base Environment Model."""
+
+    name: str = Field(nullable=False, index=True)
```
brenkao, line 19: is this prod staging and dev? could we use an enum here?

lilypad/ee/server/models/deployments.py
```diff
@@ -0,0 +1,39 @@
...
+    revision: int = Field(default=1)
```
brenkao, line 30: How does revision relate to version_num?

lilypad/ee/server/models/environments.py
koxudaxi: I also thought about that, but I believe environment names should be left to the users' discretion. Some projects might have 'testing' environments, and I've worked at companies with multiple dev environments. It's difficult to anticipate all possible environment names and how they should be handled. However, there might be benefits to specifying a default or production environment.

lilypad/ee/server/models/deployments.py
koxudaxi: There is no particular difference, so let's use version_num.

lilypad/ee/server/_utils/environment.py
```diff
@@ -0,0 +1,63 @@
...
+        return get_local_environment(session, project_uuid)
+    if api_key:
+        api_key_row = session.exec(
+            select(APIKeyTable).where(APIKeyTable.key_hash == api_key)
+        ).first()
+        if not api_key_row:
+            raise HTTPException(
+                status_code=status.HTTP_401_UNAUTHORIZED, detail="Invalid user"
+            )
+        return Environment.model_validate(api_key_row.environment)
```
koxudaxi, line 59: @brenkao @willbakst 
After thinking about it today, I realized that environments like "prd" and "stg" are fundamental and critical concepts in our system.
Initially, I considered adding LILYPAD_ENV in the client setup, but I've come to believe that API keys should be issued per environment instead. This approach would help prevent incidents caused by configuration mistakes and address potential security concerns.

This is quite a conceptual change, so I'd appreciate your review on this approach.

lilypad/ee/server/models/deployments.py
koxudaxi: OK, I have deleted it.

lilypad/ee/server/models/environments.py
brenkao: @willbakst is leaving the user to create their own production environment fine?

lilypad/ee/server/models/deployments.py
brenkao: So this version_num will be separate from the one we use in generations? I'm trying to understand the use case here

lilypad/ee/server/models/environments.py
brenkao: i guess this is really interesting to allow users to A/B test their prompts

willbakst: hmm there's a lot to think about here

if i set the environment to production and run a generation, will it automatically get versioned and added? will that version exist in other environments?

i feel like we need to have some form of fixed dev -> prod flow and then we could also allow arbitrary environments too?

willbakst: maybe this is easily solved with sync tooling to sync environments and then it doesn't matter?

migrations/versions/0019_add_environment_and_deployment_tables.py
brenkao, line 1: with the most recent pr, this will need to be updated to 0020, and the other one 0021

lilypad/ee/server/models/deployments.py
koxudaxi: In my scenario, I envision users being able to see a deployment history in the UI.
For example, if we display dates and deployment history, users can see what prompts were used at specific points in the past.
This could also be useful when someone wants to revert to a previous deployment state.Retry

lilypad/ee/server/models/environments.py
koxudaxi: I agree we need standard way for new users.
```python
class EnvironmentType(int, enum.Enum):
    """Environment types enum."""
    dev = 0
    staging = 1
    production = 2
    custom = 3

class EnvironmentBase(SQLModel):
    """Base Environment Model."""
    name: str = Field(nullable=False, index=True)
    description: str | None = Field(default=None)
    type: EnvironmentType | None = Field(default=None, sa_column=Column(Enum(EnvironmentType)))
    project_uuid: UUID = Field(
        index=True, foreign_key=f"{PROJECT_TABLE_NAME}.uuid", ondelete="CASCADE"
    )
```

This gives clear structure for standard environments while keeping custom option. The ordered enum values make dev -> staging -> production flow more explicit. How does this look?

koxudaxi: I've organized my thoughts after considering everyone's input.
I'm assuming we'll manage deployments using a web interface or CLI tool. With this approach, I don't think we need many constraints at the database table level.
Environments would be managed through UI/API, allowing users to define their own environment names while providing guidance on standard flows (dev→staging→prod). The sync tooling mentioned would help users promote deployments between environments.
This gives flexibility while still providing structure through the management interface rather than strict database constraints.
From an operational perspective, we should consider different deployment scenarios - some users might edit prompts directly in the UI before deploying (similar to existing cloud providers), while others may want to control deployments programmatically through CI pipelines.
For the managed=True case, what's the most intuitive behavior when no deployment exists in the target environment? Should it reference a deployment from a lower environment (e.g., dev/staging for production)? Or just use the latest version? Users might feel uncomfortable if undeployed items are automatically used, so perhaps an "auto deploy" option in the UI would make sense.

koxudaxi: Is the "sync tooling" you mentioned similar to what I'm thinking about? I'm envisioning a web interface or CLI tool to manage deployments between environments, allowing users to promote deployments while maintaining flexibility in environment naming.

lilypad/ee/server/models/deployments.py
koxudaxi: In other words, it is a history of deployments in the same environment

lilypad/ee/server/models/deployments.py
```diff
@@ -0,0 +1,65 @@
...
+    __table_args__ = (
+        Index(
+            "ux_environment_active_deployment",
+            "environment_uuid",
+            unique=True,
+            postgresql_where=text(
+                "is_active = true"
+            ),  # Use text() for partial index condition
+        ),
+    )
```
koxudaxi, line 62: There can only be one is_active record per environment at any given time

lilypad/ee/server/models/deployments.py
brenkao: but why do we need this version_num, because we have a connection to the generation table which has the verison_num. we can query by `deployment.generation.version_num`

koxudaxi: That's the order in which the generations were generated, right?
It means that it is necessary to take a history of deployments per environment.
Maybe they deployed generations 1 and 3 and then deployed 2.

lilypad/server/models/api_keys.py
```diff
@@ -29,6 +35,12 @@ class APIKeyBase(SQLModel):
     project_uuid: UUID = Field(
         index=True, foreign_key=f"{PROJECT_TABLE_NAME}.uuid", ondelete="CASCADE"
     )
+    environment_uuid: UUID | None = Field(
```
brenkao, line 38: should this be required? we should have a default environment on creation called  Default, they can rename it whenever. If so, when user is created, we should also create an environment for them

lilypad/server/models/api_keys.py
```diff
@@ -29,6 +35,12 @@ class APIKeyBase(SQLModel):
...
+        default=None,
+        foreign_key=f"{ENVIRONMENT_TABLE_NAME}.uuid",
+        ondelete="SET NULL",
```
brenkao, line 41: whats the idea behind using set null? if user deletes their environment, would user have to select a different environment for this api key?

lilypad/ee/server/models/deployments.py
brenkao: okay so this is mostly for rollback purposes. so they can be flexible and rollback from v1 to v3 to v2.

migrations/versions/0019_add_environment_and_deployment_tables.py
brenkao: oops once more we have another 0020. so this will need to be 0021 and 0022

koxudaxi: I have temporarily put on hold the removal of is_default because it affects the UI side as well.

lilypad/server/models/api_keys.py
koxudaxi: From a risk perspective, warn users that the api_key will be invalidated at the stage of environment removal, and
The api_key table should be deleted along with the environment.

koxudaxi: By the way, are all records physical deletions? Should I also consider logical deletion for the logs?
This is not a blocker but something to consider in the future.

lilypad/server/models/api_keys.py
koxudaxi: I updated the tables

migrations/versions/0019_add_environment_and_deployment_tables.py
brenkao: Ok I will make the change when this Pr is merged in for the client

lilypad/server/models/api_keys.py
brenkao: The way I've been doing it is adding an archived column which is a datetime field. we can do that as well.

brenkao (approved):

PR #311, opened 2025-03-10
feat: self-deploy lilypad, add license

### LINKED ISSUE #300: Make it easier to set an enterprise license (self-host) and make expiration more clear with warnings
### Description

Right now the only way to set an enterprise license if manually, and if a license expires we simply treat the user as though they are on the free tier.

We should do the following:
1. Enable setting an enterprise license in the UI from the organization's settings page
2. Warn the user that their license is expiring soon (~15 days before expiration?)
3. Don't warn if on auto-renewal (once we offer that)
4. Explain that a previously accessible feature is no longer accessible due to expiration of license for some period of time (~15 days after expiration?)

brenkao:
Closes #300


lilypad/ee/server/require_license.py
```diff
@@ -114,6 +114,12 @@ async def __call__(
                     detail="License key does not match organization",
                 )
 
+            if license_info.is_expired:
+                raise HTTPException(
```
willbakst, line 118: shouldn't this stuff be Lilypad specific exceptions? or no?

willbakst (approved):

lilypad/ee/server/require_license.py
brenkao: no, since generations creation is not under license.

willbakst: I was thinking more like managed generations, but that should actually fail.

willbakst (approved):

PR #315, opened 2025-03-14
feat: remove span order and improved trace type in ui

### LINKED ISSUE #277: Show `span.metadata` and `span.log` and other span items pretty in the UI
### Description

Right now it can be tough to find sometimes inside the full data object. Would be nice for known methods of `span` to show up pretty in the UI

Image

For example, I logged `response` which you can see in `attributes`, but would be nice to have something like a Metadata card that's pretty JSON.

Also now that I'm writing this, I'm noticing that the traces/spans are in the wrong order? Should be first at the top, last at the bottom right?

Also noticing that the token count stuff saying unknown is a little weird. We should probably just not show it?

brenkao:
Closes #277 

Removed `span.order`. The server handles order by building a tree, so span can be received in any order.


tests/test_spans.py
```diff
@@ -51,6 +52,21 @@ def get_span_context(self) -> DummySpanContext:
         """Return a dummy span context."""
         return DummySpanContext()
 
+    def set_status(self, status: Any) -> None:
+        """Set the status of the span."""
+        self.status = status
+
+    def record_exception(self, exc: BaseException) -> None:
```
brenkao, line 59: @koxudaxi 
Need your guidance on this part. I am using opentelemetry's `record_exception` so for DummySpan, I'm not sure how we should test this. See comment below

tests/test_spans.py
```diff
@@ -130,8 +145,6 @@ def test_exception_captures_stacktrace() -> None:
     assert "exception.type" in attrs
     assert "exception.message" in attrs
     assert "exception.stacktrace" in attrs
-    assert "ValueError" in attrs["exception.stacktrace"]
```
brenkao, line 133: How should I test this using `DummySpan`, We should probably use our `Span` instead right? Otherwise I'm just writing a `record_exception` function and testing it, instead of our actual `Span`

migrations/versions/0021_add_default_environment.py
brenkao, line 1: because production and other environments have api keys already, i'm temporarily removing the required constraint on environment_uuid, until we implement a default environment that we can map API keys to

tests/test_spans.py
koxudaxi: OK, I have added the testcase for the method.

koxudaxi (approved):

PR #330, opened 2025-03-18
feat: add secret-manager

koxudaxi:


koxudaxi:
The PR uses supabase postgres feature for encrypting secret. we should test it on local and staging.

lilypad/server/_utils/auth.py
```diff
@@ -143,10 +151,12 @@ async def get_current_user(
     session: Annotated[Session, Depends(get_session)],
     settings: Annotated[Settings, Depends(get_settings)],
 ) -> UserPublic:
-    """Dependency to get the current authenticated user from session."""
-    """Get the current user from JWT token"""
+    """Get current user with default scopes."""
     if token == LOCAL_TOKEN:
```
brenkao, line 155: Mostly for reference. This was mainly for `lilypad local` which we do not use anymore, so this code will eventually be wiped.

lilypad/server/_utils/auth.py
```diff
@@ -182,3 +194,28 @@ async def get_current_user(
             detail=str(e),
         )
     raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED, detail="Invalid user")
+
+
+def require_scopes(
```
brenkao, line 199: I assume this will need a lilypad frontend change to modify scopes?

scripts/migration_keys.py
```diff
@@ -0,0 +1,190 @@
...
+            session.execute(text("ALTER SYSTEM SET statement_log = 'none'"))
+            session.commit()
+            migration_logger.info("Disabled system-wide statement logging for security")
+
+        # Get all users
+        users = session.exec(select(UserTable)).all()
+        migration_logger.info(f"Found {len(users)} users to process")
+
+        # Create services
+        secret_manager = get_secret_manager(session)
```
brenkao, line 40: What should we do for self-host. This assumes if somebody self hosts lilypad, they must use supabase vault?

brenkao: note: I am commenting on all instances of `get_secret_manager`, not this specific instance

lilypad/server/_utils/auth.py
koxudaxi: I think we don't need any changes. I will check it my local.
but, we can use the scope like "admin:write" on frontend in the feature.

scripts/migration_keys.py
koxudaxi: I haven't tried it yet, but this PR for self-hosted
I changed the postgres image in docker-compose.yml to the supabase image.
We will also ask leading self-hosted users and can write implementations of other secret managers such as aws and gcp as soon as possible.

lilypad/server/_utils/auth.py
koxudaxi: we should remove it. But, we should create another pr to remove the local tokens.

brenkao: ok in a different PR i removed this

brenkao (approved):

PR #342, opened 2025-03-21
chore: rearranged api

brenkao:


lilypad/ee/server/api/v0/organization_invites_api.py
willbakst, line 1: if we're not restricting users for self-host, does this need to be in EE?

lilypad/ee/server/models/organization_invites.py
willbakst, line 1: same thing here, cloud user can't change code and we don't care to block this for self-host since it's easy to implement

lilypad/ee/server/models/user_organizations.py
willbakst, line 1: ditto

lilypad/ee/server/schemas/deployments.py
willbakst, line 1: to confirm my understanding, deployments are really only available when using playground generations which is why this is in ee right?

lilypad/ee/server/api/v0/user_organizations_api.py
```diff
@@ -0,0 +1,144 @@
...
+
+from ee.validate import LicenseValidator, Tier
+
+from .....server._utils import get_current_user
+from .....server.models import UserTable
+from .....server.schemas import (
+    UserPublic,
+)
+from .....server.services import OrganizationService
+from ...features import cloud_features
```
willbakst, line 18: am i crazy that I can't find this?

lilypad/ee/server/api/v0/user_organizations_api.py
```diff
@@ -0,0 +1,144 @@
...
+    license_info = validator.validate_license(
+        org_invite.organization_uuid, organization_service
+    )
+    tier = Tier.FREE
+    if license_info:
+        tier = license_info.tier
+    num_users = user_organization_service.count_users_in_organization(
+        org_invite.organization_uuid
+    )
+    if num_users >= cloud_features[tier].num_users_per_organization:
```
willbakst, line 92: can't really confirm what's happening here cause I can't find `cloud_features` for whatever reason, but this feels like it's only part of what we need? don't we also need to ensure the "as licensed" limit for self host when there's a license present?

lilypad/ee/server/schemas/deployments.py
brenkao: this will be moved out of ee in a different pr.

lilypad/ee/server/api/v0/user_organizations_api.py
brenkao: its in lilypad/server/features

lilypad/ee/server/api/v0/user_organizations_api.py
brenkao: we need to update this to check license.num_users_per_organization. when we update the license code we can update this line

lilypad/ee/server/models/user_organizations.py
brenkao: since we have code in user_organizations under license, everything in user_organizations is in ee.

lilypad/ee/server/api/v0/organization_invites_api.py
brenkao: hmm maybe not, I can move organization_invites back

lilypad/ee/server/models/organization_invites.py
brenkao: same as above

willbakst (approved):

PR #347, opened 2025-03-25
feat: preparation for mirascope v2.0

brenkao:
* Removed all sdk code except what was necessary for server
* Renamed generation to function


lilypad/server/_utils/versions.py
```diff
@@ -11,7 +11,7 @@ def construct_function(
     import_line = "import lilypad" if include_import else ""
     func_def = f"""
     {import_line}
-    @lilypad.generation(managed=True)
+    @lilypad.function(managed=True)
     def {function_name}({", ".join(arg_list)}) -> lilypad.Message: ...
```
willbakst, line 15: nit: this will no longer work as expected, but I figure you'll tackle this in the playground pr and this is just find/replace

willbakst (approved):

lilypad/server/_utils/versions.py
brenkao: yeah this is just find and replace

PR #350, opened 2025-03-27
fix: fix client access from playground

koxudaxi:


brenkao (approved):

brenkao:
I updated debug statement back to old logic

PR #375, opened 2025-04-03
fix: support all mirascope providers in playground

koxudaxi:


playground-requirements.txt
```diff
@@ -1 +1,2 @@
-lilypad-sdk[openai,anthropic,azure,bedrock,google,mistral,outlines]@git+https://github.com/Mirascope/lilypad-sdk-python.git@main
\ No newline at end of file
+lilypad-sdk[openai,anthropic,azure,bedrock,google,mistral]@git+https://github.com/Mirascope/lilypad-sdk-python.git@main
```
willbakst, line 1: shouldn't we fix the version?

koxudaxi: OK, I fixed the version.

brenkao (approved):

PR #378, opened 2025-04-04
feat: privacy and tos

brenkao:


koxudaxi (approved):

client/src/components/GoogleLogin.tsx
```diff
@@ -117,6 +117,8 @@ export const GoogleLogin = ({ redirect }: { redirect?: string }) => {
     });
     const stateObject = {
       provider: "google",
+      privacy_version: "2.0",
```
willbakst, line 120: no version, just the last updated date that was accepted, and we'll update the date when we update

lilypad/server/models/user_consents.py
```diff
@@ -0,0 +1,45 @@
...
+from sqlmodel import Field, Relationship, SQLModel
+
+from .base_sql_model import BaseSQLModel
+from .table_names import USER_CONSENT_TABLE_NAME, USER_TABLE_NAME
+
+if TYPE_CHECKING:
+    from .users import UserTable
+
+
+class UserConsentBase(SQLModel):
```
willbakst, line 18: same here, update to the last updated date that was accepted

client/src/routes/auth.login.tsx
```diff
@@ -38,42 +35,40 @@ export const Route = createFileRoute("/auth/login")({
...
-            </>
-          )}
+          <GithubLogin redirect={redirect} />
+          <GoogleLogin redirect={redirect} />
         </CardContent>
+        <CardFooter>
+          <Typography variant='p' affects='muted'>
+            By signing in, you agree to our{" "}
+            <a
+              href='https://mirascope.com/terms'
```
willbakst, line 54: this will be mirascope.com/terms/service when we deploy

PR #388, opened 2025-04-09
feat: updated privacy and tos to be a popup

brenkao:
Note: this relies on terms and privacy to be stable


client/src/utils/auth.ts
```diff
@@ -1,55 +1,82 @@
...
+        return await fetchGoogleCallbackCode(code);
       } else {
         return null;
       }
     },
     enabled: Boolean(code),
   });
+
+export const fetchVersions = async () => {
+  const privacyResponse = await fetch("https://beta.mirascope.com/privacy"); // Adjust URL to your actual privacy policy page
```
willbakst, line 37: these will be mirascope.com on launch, so let's update these to that

willbakst (approved):

willbakst (approved):

PR #405, opened 2025-04-14
feat: add backend APIs for annotation assignment

koxudaxi:
Related Issues: https://github.com/Mirascope/lilypad-sdk-python/issues/72


lilypad/ee/server/api/v0/spans_api.py
```diff
@@ -0,0 +1,126 @@
...
+    summary="Assign Annotation to User",
+    description="Assigns the annotation associated with the given span to a user specified by email. If no annotation exists, it will be created.",
+    status_code=status.HTTP_200_OK,
+    responses={
+        status.HTTP_404_NOT_FOUND: {"description": "Span or User not found"},
+        status.HTTP_400_BAD_REQUEST: {"description": "Invalid input or state"},
+        status.HTTP_500_INTERNAL_SERVER_ERROR: {"description": "Internal server error"},
+    },
+)
+@require_license(tier=Tier.ENTERPRISE, cloud_free=True)
```
brenkao, line 31: why is this EE?

lilypad/ee/server/api/v0/spans_api.py
```diff
@@ -0,0 +1,126 @@
...
+async def assign_span_annotation(
+    span_uuid: UUID,
+    assignment_data: SpanAssignmentUpdate,
+    span_service: Annotated[SpanService, Depends(SpanService)],
+    annotation_service: Annotated[AnnotationService, Depends(AnnotationService)],
+    user_service: Annotated[UserService, Depends(UserService)],
+    current_user: Annotated[UserPublic, Depends(get_current_user)],
+) -> list[AnnotationPublic]:
+    """Handles assigning the annotation for a span_uuid to a user by email."""
+    organization_uuids_to_search = [
```
brenkao, line 41: i think we can simplify this by using the organization_uuid that is associated to the project. grab the project and its organization uuid, then we can search for users in that organization

lilypad/ee/server/api/v0/spans_api.py
```diff
@@ -0,0 +1,126 @@
...
+from lilypad.ee.server.schemas.spans import SpanAssignmentUpdate
+from lilypad.ee.server.services import AnnotationService
+from lilypad.server._utils import get_current_user
+from lilypad.server.schemas import SpanMoreDetails, UserPublic
+from lilypad.server.services import SpanService, UserService
+
+spans_router = APIRouter()
+
+
+@spans_router.patch(
```
brenkao, line 19: To me this is a bit strange that it is in spans. The way that "assignment" works is that we can create an empty Annotation. so it should just be a post annotation endpoint. 

This way, it makes more sense, since annotations is an EE feature for self host.

We should be able to create an AnnotationCreate object in the SDK and pass in the span_uuid from there

lilypad/server/services/users.py
```diff
@@ -49,3 +50,19 @@ def update_user_keys(self, data: dict) -> UserTable:
         self.session.flush()
         self.session.refresh(user)
         return user
+
+    def find_record_by_email_in_organizations(
```
brenkao, line 54: I dont think we need this, we should be able to select user_organizations from organization_uuid. then we can access the user by `user_organizations.user`

This does make a lazy second database query, but we should be able to update our `UserOrganization` model and change the relationship to always join user if performance is an issue.

lilypad/ee/server/api/v0/spans_api.py
koxudaxi: Is it always empty? If the annotation has already been created, does it raise an error?

brenkao: Right now I dont put any restriction on annotation created. We can add a constraint that if span_uuid and assigned_to already exist, then we raise a duplicate error.

lilypad/ee/server/schemas/annotations.py
```diff
@@ -34,6 +34,22 @@ class AnnotationCreate(BaseModel):
     type: EvaluationType | None = None
     data: dict[str, Any] | None = None
     assigned_to: list[UUID] | None = None
+    assignee_email: str | None = None
```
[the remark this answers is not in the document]
koxudaxi: > if its the latter, should we make assigned to email a list of emails?
yes this one. make sense. I will change the type.

brenkao (approved):

PR #406, opened 2025-04-14
feat: Add tags_name to span_update

koxudaxi:
Related Issues: https://github.com/Mirascope/lilypad-sdk-python/issues/64


willbakst:
Why are we adding a link between functions and tags?

The idea here is that it's a tag for the trace, right? e.g. `trace.tag(...)`

koxudaxi:
> Why are we adding a link between functions and tags?
> 
> The idea here is that it's a tag for the trace, right? e.g. `trace.tag(...)`

Yes. When this function is executed, the linked Tag Table is expected to ultimately be linked to a record on the Span Table. Should we link the tag during Span registration by searching for the tag, rather than going through the Function?

willbakst:
> > Why are we adding a link between functions and tags?
> > The idea here is that it's a tag for the trace, right? e.g. `trace.tag(...)`
> 
> Yes. When this function is executed, the linked Tag Table is expected to ultimately be linked to a record on the Span Table. Should we link the tag during Span registration by searching for the tag, rather than going through the Function?

Yes, we shouldn't go through function

brenkao:
> > Why are we adding a link between functions and tags?
> > The idea here is that it's a tag for the trace, right? e.g. `trace.tag(...)`
> 
> Yes. When this function is executed, the linked Tag Table is expected to ultimately be linked to a record on the Span Table. Should we link the tag during Span registration by searching for the tag, rather than going through the Function?

If I'm understanding correctly, we want to be able to tag spans from the SDK? I feel like we have two options.

1. Create the tags if they dont exist / map tags to span if they do exist in the traces_api.
2. In the SDK, we pull the tags first, then make a call to the create_tag api using the SDK before we trace. Then we trace, then we update the span tag link.

Option 1 seems preferred

brenkao:
We should not need to make any schema changes for this issue

koxudaxi:
@brenkao 
> If I'm understanding correctly, we want to be able to tag spans from the SDK? I feel like we have two options.

Does that mean you always call the API twice? Three times when creating?
I was thinking of accessing the DB within spanAPI, as this might affect performance, but what do you think?

lilypad/server/schemas/spans.py
```diff
@@ -521,7 +521,15 @@ class SpanCreate(SpanBase):
 class SpanUpdate(BaseModel):
     """Span update model"""
 
-    tags: list[TagPublic] | None = None
+    tags_by_uuid: list[UUID] | None = None
+    tags_by_name: list[str] | None = None
```
koxudaxi, line 525: Now, the api supports tags_uuid and tags_name for SDK and frontend.

koxudaxi:
> Does that mean you always call the API twice? Three times when creating? I was thinking of accessing the DB within spanAPI, as this might affect performance, but what do you think?
> 
> The same reason I created FunctionTagLink.

Now, the api supports tags_uuid and tags_name for SDK and frontend.

brenkao (approved):

PR #416, opened 2025-04-16
fix: latency

brenkao:
Potentially fix latency


willbakst (approved):

koxudaxi (approved):
we should use Supavisor(connections pooler)  https://supabase.com/docs/guides/troubleshooting/how-to-change-max-database-connections-_BQ8P5

brenkao:
> we should use Supavisor(connections pooler) https://supabase.com/docs/guides/troubleshooting/how-to-change-max-database-connections-_BQ8P5

When we hit db issues again, we can revisit this

PR #426, opened 2025-04-21
feat: implement search

brenkao:


client/src/routes/_auth/projects/$projectUuid/traces.$.tsx
```diff
@@ -34,8 +35,31 @@ export const Trace = () => {
 
 export const TraceBody = () => {
   const { projectUuid, _splat: traceUuid } = useParams({ from: Route.id });
-  const { data } = useSuspenseQuery(tracesQueryOptions(projectUuid));
+  const { data: defaultData } = useSuspenseQuery(
+    tracesQueryOptions(projectUuid)
+  );
```
koxudaxi, line 40: What is the reason for outputting the contents of the traditional DB instead of opensearch in the first display?

brenkao: Opensearch may drop spans. Unless we can guarantee 100% uptime, if opensearch goes down, we still show their traces and lilypad still works. search will just be unavailable.

willbakst (approved):

willbakst (approved):

PR #429, opened 2025-04-22
feat: Refactor trace handling to support infinite scrolling.

koxudaxi:


brenkao:
Functions also need to be updated. `CompareTracesTable.tsx` and `FunctionSpans.tsx`, best would be to create a `use-pagination` hook so we can share the logic.

Also when #426 gets merged in, we also need to paginate that as well.

koxudaxi:
@brenkao 
OK, I have udpated the components.

client/src/hooks/usePaginatedQuery.tsx
brenkao, line 1: for hooks, lets use kebab case.

client/src/components/CompareTracesTable.tsx
```diff
@@ -11,13 +11,34 @@ export const CompareTracesTable = ({
...
-        ...spansByFunctionQueryOptions(projectUuid, uuid),
-      })),
-  });
-  const flattenedData = data.map((result) => result.data).flat();
-  return <TracesTable data={flattenedData} />;
-};
+  const first = usePaginatedSpansByFunction(projectUuid, firstFunctionUuid);
+  const second = usePaginatedSpansByFunction(
+    projectUuid,
+    secondFunctionUuid ?? "__none__",
```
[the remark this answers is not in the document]
koxudaxi: I updated the request.

client/src/components/DataTable.tsx
```diff
@@ -69,290 +66,318 @@ interface GenericDataTableProps<T> {
...
             </Suspense>
           </ResizablePanel>
         </>
       )}
     </ResizablePanelGroup>
   );
-};
+}
+
+export const DataTable = forwardRef(DataTableInner) as
```
brenkao, line 380: not sure this is necessary, react 19 has built-in ref so the original DataTable should have ref without needing to forwardRef (which is being deprecated)

client/src/components/DataTable.tsx
brenkao, line 1: Something about this PR broke selecting rows, when checking the checkbox in the header, and individual checkboxes the annotation button should work and compare should work

brenkao, line 1: Sorting is also a little weird. When you sort by timestamp, it only sorts by the loaded data.

client/src/components/DataTable.tsx
```diff
@@ -69,290 +66,318 @@ interface GenericDataTableProps<T> {
...
-                      className='h-24 text-center'
-                    >
-                      No results.
-                    </TableCell>
-                  </TableRow>
-                )}
-              </TableBody>
-            </Table>
-          </div>
+        <table className="w-full text-sm border-separate border-spacing-0">
```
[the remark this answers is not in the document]
brenkao: As a result,the original styling is missing and the labels don't align with the data

client/src/components/TracesTable.tsx
```diff
@@ -230,59 +232,104 @@ const DetailPanel = ({ data, path }: { data: SpanPublic; path?: string }) => {
...
+      setSelectedRows([...selectedRows, row]);
+    } else {
+      setSelectedRows(
+        selectedRows.filter((item) => item.span_id !== row.span_id)
+      );
+    }
+  };
+  
+  const sentinelRef = useRef<HTMLTableRowElement>(null);
+  console.log(data)
```
brenkao, line 317: remove

client/src/utils/spans.ts
```diff
@@ -233,3 +234,20 @@ export const spansSearchQueryOptions = (
...
+  limit: number;
+}
+
+export const fetchSpansByFunctionUuidPaged = async (
+  projectUuid: string,
+  functionUuid: string,
+  { offset, limit }: PageParam,
+) => (
+  await api.get<PaginatedSpanPublic>(
+    `/projects/${projectUuid}/functions/${functionUuid}/spans/paginated` +
```
brenkao, line 249: best to use a urlsearchparam to construct limit and offet.

client/src/components/DataTable.tsx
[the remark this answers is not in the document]
koxudaxi: I have tried everything, but if there is a sticky, it doesn't actually become sticky, but scrolls to after the table.
In other words, the column headers are not sticky.
Is there a better way?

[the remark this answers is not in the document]
koxudaxi: OK I have fixed it

client/src/components/DataTable.tsx
koxudaxi: If I drop `forwardRef`, CI turns red because TypeScript can’t yet treat a **generic** function component with a `ref` as valid:

```
TS2786: 'DataTable' cannot be used as a JSX component
TS2322: Property 'ref' does not exist …
```

React 19 runtime is fine, but `@types/react` hasn't caught up.  
Until those typings land, we need the `forwardRef` wrapper.
https://github.com/Mirascope/lilypad/actions/runs/14672639582/job/41182528960?pr=429#step:7:119

koxudaxi:
OK, I fixed all things

client/src/components/DataTable.tsx
brenkao: got it okay

brenkao (approved):

PR #467, opened 2025-05-09
feat: Implement Stripe span metering for usage tracking

koxudaxi:
Related PR: https://github.com/Mirascope/lilypad-sdk-python/pull/127


lilypad/server/api/v0/auth/utils.py
```diff
@@ -15,12 +17,20 @@ def handle_user(
...
     posthog: posthog.Posthog,
+    billing_service: BillingService
 ) -> UserPublic:
     """Handle user creation or retrieval."""
     user = session.exec(select(UserTable).where(UserTable.email == email)).first()
 
     if user:
         user_public = UserPublic.model_validate(user)
+
+        if user_public.active_organization_uuid:
```
willbakst, line 28: how do we handle the self-host case?

lilypad/server/models/organizations.py
```diff
@@ -37,5 +38,8 @@ class OrganizationTable(OrganizationBase, BaseSQLModel, table=True):
     tags: list["TagTable"] = Relationship(
         back_populates="organization", cascade_delete=True
     )
+    billings: list["BillingTable"] = Relationship(
```
willbakst, line 41: can you explain why this is a list please?

koxudaxi: nice catch.it  should be a single record.

lilypad/server/api/v0/auth/utils.py
koxudaxi: Indeed. Should we ignore the self-hosted case for now?

willbakst: We need to handle the self-hosted case, in which case there will be a license (like before) and no stripe billing.

koxudaxi: I added the validation for self-hosted license.

willbakst (approved):

PR #498, opened 2025-05-19
feat: add mirascope view for lilypad panels

brenkao:


lilypad/server/schemas/span_more_details.py
```diff
@@ -125,7 +125,8 @@ def convert_gemini_messages(
             attribute_message = json.loads(attributes.get("message", "{}"))
             if content := attribute_message.get("content"):
                 for c in content:
```
willbakst, line 127: nit: this should be `for part in content` since `c` is a part

willbakst (approved):

PR #508, opened 2025-05-28
fix: improve code panel layout for better scrolling behavior

koxudaxi:


brenkao:
I know this is draft but any updates to UI components should be made in `mirascope/ui`. Create a storybook that showcases the error and the solution to fix it. We may just want to pass in className since it might not make sense for every component to have the flex

brenkao (approved):

PR #516, opened 2025-06-03
feat: add Kafka topic setup and recent spans polling endpoint

koxudaxi:


sdks/python/src/lilypad/_configure.py
```diff
@@ -85,104 +63,41 @@ def __init__(self) -> None:
...
-            finally:
-                self._q.task_done()
+            return SpanExportResult.FAILURE
+        except Exception as exc:
+            self.log.error("Unexpected error sending spans: %s", exc)
+            return SpanExportResult.FAILURE
+
+        self.log.debug(f"Spans {response.trace_status}: {response.span_count} spans")
+        if response.trace_status == "queued" and response.span_count > 0:
+            # When using Kafka queue, we don't get database UUIDs back immediately
```
brenkao, line 87: We should get some `span_uuid` from the initial add to queue, so shouldn't we be able to get some uuids back? If not, then even returning back the trace_id would be really useful, and I can always render the client using trace_id if we dont have uuid.

app/lilypad/server/services/kafka.py
```diff
@@ -0,0 +1,207 @@
...
+                    retry_delay *= 2  # Exponential backoff
+                else:
+                    logger.error(
+                        f"Failed to initialize Kafka producer after {max_retries} attempts: {e}"
+                    )
+                    return False
+
+        return False
+
+    async def send_span(self, span_data: dict[str, Any], user_id: UUID) -> bool:
```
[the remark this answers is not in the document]
koxudaxi: > Our KafkaService can depend on get_current_user so we don't need to pass in user_id.

Unfortunately, KafkaService is singleton. Although it is not impossible to sandwich an even thinner wrapper...

sdks/python/src/lilypad/_configure.py
koxudaxi: OK, I added the trace_id in the url.

app/client/src/routes/_auth/projects/$projectUuid/traces.$.tsx
brenkao, line 1: I am confused about the real-time polling. We already have polling so what is this trying to accomplish?

Note: I keep getting `Real-time updates stopped due to errors`, is there any configuration I need to do?

koxudaxi: > I am confused about the real-time polling. We already have polling so what is this trying to accomplish?

I don't understand it? Where is the implementation part?

> Note: I keep getting Real-time updates stopped due to errors, is there any configuration I need to do?

No need additonal configuration. But, we need the backend api.

koxudaxi: OK, I found the implementation and re-used it.

brenkao: Yeah I would not worry too much about the frontend. We can always add the real-time update as a separate feature. I want to get all the infra stuff out and stable.

app/migrations/versions/0040_add_span_polling_indexes.py
brenkao, line 1: Since unique-organization-uuid pr has been merged, this will need to be bumped up 1 version

brenkao (approved):

app/migrations/versions/0040_add_span_polling_indexes.py
koxudaxi: I fixed it.

PR #528, opened 2025-06-10
Add distributed tracing support

koxudaxi:


sdks/python/docs/distributed-tracing.md
```diff
@@ -0,0 +1,342 @@
+# Distributed Tracing Guide
+
+Lilypad supports distributed tracing using OpenTelemetry context propagation, allowing you to trace requests across multiple services and processes.
+
+## Important Note on Global State
+
+**WARNING**: Lilypad's context propagation modifies the global OpenTelemetry propagator by calling `propagate.set_global_textmap()`. This affects all OpenTelemetry instrumentation in your process. If you have other OpenTelemetry integrations, they will use Lilypad's configured propagator.
```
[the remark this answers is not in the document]
brenkao: They would have to pass in an exporter to lilypad.configure that overwrites our exporter, or we append to the exporter that they pass in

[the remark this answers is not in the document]
koxudaxi: OpenTelemetry supports multiple exporters through the CompositePropagator mechanism. According to the https://opentelemetry.io/docs/specs/otel/context/api-propagators/#composite-propagator
```py
from opentelemetry.propagators.composite import CompositePropagator

# Combine multiple propagators
composite = CompositePropagator([
    TraceContextTextMapPropagator(),
    B3Format(),
    # Existing user propagators...
])
```

https://opentelemetry-python.readthedocs.io/en/latest/api/propagators.composite.html#opentelemetry.propagators.composite.CompositePropagator
We can store the existing propagator configuration before applying Lilypad's configuration, allowing users to maintain their current setup.

sdks/python/pyproject.toml
```diff
@@ -32,6 +32,8 @@ dependencies = [
     "opentelemetry-api>=1.31.0",
     "opentelemetry-sdk>=1.31.0",
     "opentelemetry-instrumentation>=0.51b0",
+    "opentelemetry-propagator-b3>=1.31.0",
```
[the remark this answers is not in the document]
koxudaxi: we do have both opentelemetry-propagator-b3 and opentelemetry-propagator-jaeger included. These packages, combined with the built-in propagators, cover all the officially supported formats.

https://opentelemetry.io/docs/languages/sdk-configuration/general/#otel_propagators

 The third-party ones like xray and ottrace would need to be installed separately by users if they need them, which seems reasonable since they're vendor-specific.


> Accepted values for OTEL_PROPAGATORS are:

> "tracecontext": [W3C Trace Context](https://www.w3.org/TR/trace-context/)
> "baggage": [W3C Baggage](https://www.w3.org/TR/baggage/)
> "b3": [B3 Single](https://opentelemetry.io/docs/specs/otel/context/api-propagators#configuration)
> "b3multi": [B3 Multi](https://opentelemetry.io/docs/specs/otel/context/api-propagators#configuration)
> "jaeger": [Jaeger](https://www.jaegertracing.io/docs/1.21/client-libraries/#propagation-format)
> "xray": [AWS X-Ray](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader) (third party)
> "ottrace": [OT Trace](https://github.com/opentracing?q=basic&type=&language=) (third party)
> "none": No automatically configured propagator.

sdks/python/docs/distributed-tracing.md
```diff
@@ -0,0 +1,342 @@
...
+
+## Manual HTTP Client Integration
+
+If you prefer explicit control or automatic instrumentation doesn't work for your use case:
+
+### Using Traced HTTP Clients
+
+```python
+# Using pre-built traced clients
+from lilypad.integrations.http_client import TracedRequestsSession
```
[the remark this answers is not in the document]
koxudaxi: Yes, you are correct. TracedRequestsSession is for the requests package specifically. It wraps requests.Session class. The naming is not clear about this connection. I will remove this from public API because automatic instrumentation already handles this case.

 I agree with your suggestion. We should make automatic instrumentation the main method. Manual methods can go to troubleshooting section only. I will remove manual examples from main documentation.

sdks/python/docs/distributed-tracing.md
```diff
@@ -0,0 +1,342 @@
...
+
+### 2. Automatic Instrumentation (Recommended)
+
+The easiest way to enable distributed tracing is using automatic instrumentation:
+
+```python
+from lilypad.integrations import instrument_http_clients
+
+# Enable at application startup
+instrument_http_clients()
```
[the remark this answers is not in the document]
koxudaxi: Make sense . I implement it.

sdks/python/src/lilypad/integrations/__init__.py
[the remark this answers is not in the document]
koxudaxi: OK, I deleted it.

sdks/python/docs/distributed-tracing.md
```diff
@@ -0,0 +1,464 @@
...
+You can also selectively instrument specific HTTP clients:
+
+```python
+import requests
+import httpx
+
+lilypad.configure(
+    api_key="your-api-key",
+    project_id="your-project-id",
+    instrument=[requests, httpx]  # Pass module objects to instrument specific clients
```
willbakst, line 85: as mentioned in my other comment, we should select a single path for instrumentation.

I think I would rather something like `lilypad.instrument_requests` over this path if we have to choose just one, which we already support as documented below

sdks/python/docs/distributed-tracing.md
```diff
@@ -0,0 +1,464 @@
...
+
+# Configure and enable automatic HTTP instrumentation
+lilypad.configure(
+    api_key="your-api-key",
+    project_id="your-project-id",
+    auto_http=True  # Enable automatic HTTP client instrumentation
+)
+
+# Now ALL HTTP calls will automatically propagate trace context!
+import requests
```
willbakst, line 67: why do we need to import requests after configure here but we don't have to do that for e.g. OpenAI and our instrumentation for that?

sdks/python/docs/distributed-tracing.md
```diff
@@ -0,0 +1,464 @@
...
+lilypad.configure(api_key="your-key", project_id="your-project")
+
+# Instrument specific clients
+lilypad.instrument_requests()
+lilypad.instrument_httpx()
+lilypad.instrument_aiohttp()
+lilypad.instrument_urllib3()
+
+# Or instrument all at once
+lilypad.instrument_http_clients()
```
willbakst, line 103: we should remove this in favor of `auto_http`. I think we should either do each one specifically or use `auto_http`

sdks/python/docs/distributed-tracing.md
```diff
@@ -0,0 +1,342 @@
...
+
+On the receiving service, extract trace context from incoming requests:
+
+```python
+from fastapi import FastAPI, Request
+from lilypad import trace
+
+app = FastAPI()
+
+@app.post("/api/process")
```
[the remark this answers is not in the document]
willbakst: To confirm, this won't be the case in this PR since we haven't instrumented FastAPI, correct?

sdks/python/docs/distributed-tracing.md
```diff
@@ -0,0 +1,464 @@
...
+
+# Consumer
+@lilypad.trace()
+def handle_message(data: dict):
+    # Process the message
+    return process(data)
+
+def process_message(message: dict):
+    # Use context manager to extract context from message headers
+    with lilypad.context(extract_from=message.get("headers", {})):
```
willbakst, line 328: what's the different between `lilypad.context` and `lilypad.propagated_context`?

sdks/python/docs/distributed-tracing.md
koxudaxi: This is due to differences in implementation.  I will fix it.

sdks/python/docs/distributed-tracing.md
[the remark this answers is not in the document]
koxudaxi: You're correct. Since we haven't implemented FastAPI instrumentation in this PR, `process_request` itself won't create a span. The `process_data` function will be traced as a child of the upstream service's span (extracted via `propagated_context`), but there won't be an intermediate span for the FastAPI endpoint itself.

I should clarify this in the documentation to avoid confusion.

sdks/python/docs/distributed-tracing.md
koxudaxi: `propagated_context` is for http request


```py
with lilypad.propagated_context(dict(request.headers)):
    # Extract trace context from HTTP headers
    process_data(data)
```

`context` is for other cases like message queue.
```py
# Cross-thread propagation
with lilypad.context(parent=saved_context):
    process_in_thread(data)

# Message queue extraction
with lilypad.context(extract_from=message["headers"]):
    handle_message(data)
```

willbakst: But aren't we propagating context in both cases to connect a parent to a child? Why have different methods?

sdks/python/docs/distributed-tracing.md
koxudaxi: OK, I deleted it.

sdks/python/docs/distributed-tracing.md
```diff
@@ -145,7 +132,7 @@ lilypad.configure(
     auto_http=True  # Enable automatic HTTP instrumentation
 )
 
-import requests  # Import after configure for auto-instrumentation
+import requests  # Can be imported before or after configure()
```
willbakst, line 135: just import it before since that's what people will do

sdks/python/docs/distributed-tracing.md
```diff
@@ -410,24 +393,25 @@ lilypad.instrument_http_clients()  # All clients
...
+```python
+# Both of these work now:
+
+# Option 1: Import before configure
+import requests
+lilypad.configure(auto_http=True)
+
+# Option 2: Import after configure
+lilypad.configure(auto_http=True)
+import requests
```
willbakst, line 413: why would anyone do this?

sdks/python/examples/auto_instrumentation_demo.py
```diff
@@ -0,0 +1,127 @@
...
+
+# Configure Lilypad with auto-instrumentation
+lilypad.configure(
+    # api_key="your-api-key",
+    # project_id="your-project-id",
+    auto_http=True  # Enable auto-instrumentation for HTTP libraries
+)
+
+# Now import HTTP libraries - they will be automatically instrumented
+import requests
```
willbakst, line 17: same thing here

sdks/python/docs/distributed-tracing.md
```diff
@@ -0,0 +1,448 @@
...
+While automatic instrumentation handles context injection, you may need to manually extract context in special cases. Use Lilypad's context managers for this:
+
+```python
+# Use Lilypad's context managers for extraction
+with lilypad.propagated_context(dict(request.headers)):
+    # Your traced functions here will be children of the incoming trace
+    process_data(data)
+
+# Or for more control
+with lilypad.context(extract_from=message["headers"]):
```
willbakst, line 251: why is this still here?

sdks/python/docs/distributed-tracing.md
koxudaxi: It's for an example for import order. I will remove the pattern.

sdks/python/docs/distributed-tracing.md
koxudaxi: For some reason, the changes are being rolled back. I will fix it right away.

koxudaxi: I changed it to `propagated_context `

sdks/python/docs/distributed-tracing.md
```diff
@@ -0,0 +1,448 @@
...
+    # Pass to worker thread
+    thread = threading.Thread(
+        target=worker_process, 
+        args=(data, current_ctx)
+    )
+    thread.start()
+
+def worker_process(data: dict, parent_ctx):
+    # Use context manager for parent context
+    with lilypad.context(parent=parent_ctx):
```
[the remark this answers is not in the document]
koxudaxi: Fixed

sdks/python/docs/distributed-tracing.md
```diff
@@ -0,0 +1,448 @@
...
+
+# Consumer
+@lilypad.trace()
+def handle_message(data: dict):
+    # Process the message
+    return process(data)
+
+def process_message(message: dict):
+    # Use context manager to extract context from message headers
+    with lilypad.context(extract_from=message.get("headers", {})):
```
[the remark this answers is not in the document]
koxudaxi: Fixed

Issue #565, opened 2025-06-23
Create DataRetentionScheduler


koxudaxi:
DataRetentionScheduler is created and started in main.py to run daily at 2 AM UTC for cloud deployments.

* Implement `DataRetentionScheduler`
*  Add scheduler initialization in lifespan function
* Ensure it only runs when `is_lilypad_cloud=True`
* Verify scheduler starts successfully on deployment

PR #569, opened 2025-06-23
feat: implement data retention service and scheduler for managing data retention policies

koxudaxi:
Cloeses: https://github.com/Mirascope/lilypad/issues/565

### DIFF
--- a/app/lilypad/ee/server/models/annotations.py
+++ b/app/lilypad/ee/server/models/annotations.py
@@ -4,6 +4,7 @@ from enum import Enum
 from typing import TYPE_CHECKING, Any
 from uuid import UUID
 
+from sqlalchemy import Index
 from sqlmodel import Field, Relationship, SQLModel
 
 from lilypad.server.models.base_sql_model import get_json_column
@@ -54,6 +55,14 @@ class AnnotationTable(AnnotationBase, BaseOrganizationSQLModel, table=True):
     """Annotation table."""
 
     __tablename__ = ANNOTATION_TABLE_NAME  # type: ignore
+    __table_args__ = (
+        # Index for cascade delete performance during data retention
+        Index(
+            "idx_annotations_span_uuid",
+            "span_uuid",
+            postgresql_using="btree",
+        ),
+    )
     project_uuid: UUID | None = Field(
         default=None, foreign_key=f"{PROJECT_TABLE_NAME}.uuid", ondelete="CASCADE"
     )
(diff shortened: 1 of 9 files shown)


willbakst:
I think you should create a new branch `LILYPAD-183` for the parent issue and merge PRs for each sub-issue into the parent branch. Then, when the parent branch is merged, the parent issue will close as well, and everything can get merged in together as a single feature rather than in pieces.

app/lilypad/server/main.py
```diff
@@ -54,6 +55,23 @@ def run_migrations() -> None:
         log.error(f"Migration failed: {e.stderr}")
 
 
+def is_lilypad_cloud_deployment() -> bool:
```
brenkao, line 58: update the original `is_lilypad_cloud` instead of creating another one

app/lilypad/ee/server/models/annotations.py
```diff
@@ -54,6 +55,14 @@ class AnnotationTable(AnnotationBase, BaseOrganizationSQLModel, table=True):
     """Annotation table."""
 
     __tablename__ = ANNOTATION_TABLE_NAME  # type: ignore
+    __table_args__ = (
```
brenkao, line 58: If we want to add an index on span_uuid, on `span_uuid` add `index=True` instead of using `table_args`. `table_args` Indexes should mostly be used for multi-column index

app/lilypad/server/services/data_retention_service.py
```diff
@@ -0,0 +1,806 @@
...
+                    log.warning(f"Serialization failure, retrying in {wait_time}s...")
+                    await asyncio.sleep(wait_time)
+                else:
+                    raise
+
+    async def _run_cleanup(self) -> None:  # pragma: no cover
+        """Run the actual cleanup process."""
+        log.info("Starting scheduled data retention cleanup")
+
+        session = create_session()
```
brenkao, line 742: I know I added `create_session`, but reviewing this. it looks difficult to manage since it's not obvious where the session is committed or rollbacked on failure.

I feel like this usage of session will lead to inevitable bugs. Would we be able to use the context manager so it can manage the cleanup for us instead?

koxudaxi: OK I have created "standalone_session" session manager

app/lilypad/ee/server/models/annotations.py
koxudaxi: I fixed it.

brenkao (approved):

PR #570, opened 2025-06-24
feat: implement data retention service and scheduler for managing data retention policies

koxudaxi:


brenkao (approved):

app/migrations/versions/0043_add_indexes_for_data_retention.py
brenkao, line 1: since other PRs are getting merged, this migration version will need to be bumped.

Issue #567, opened 2025-06-23
Implement display retention filtering in SpanService


koxudaxi:
Add display-level date filtering to all span/trace retrieval methods based on user tier.

* Create `_apply_display_retention_filter()` method
* Apply to: `find_all_no_parent_spans()`, `find_records_by_function_uuid_paged()`
* Free: 30-day display, Pro: 180-day display, Team/Enterprise: unlimited

PR #572, opened 2025-06-24
feat: refactor organization tier retrieval and apply display retention

koxudaxi:
Closes: https://github.com/Mirascope/lilypad/issues/567

### DIFF
--- a/app/lilypad/server/services/billing.py
+++ b/app/lilypad/server/services/billing.py
@@ -5,10 +5,11 @@ import time
 import uuid
 from datetime import datetime, timezone
 from typing import Any
+from uuid import UUID
 
 import stripe
 from fastapi import HTTPException, status
-from sqlmodel import Session, desc, select, update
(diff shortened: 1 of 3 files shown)


app/lilypad/server/_utils/tier.py
```diff
@@ -0,0 +1,82 @@
...
+
+        if not billing:
+            return Tier.FREE
+
+        if not billing.stripe_price_id:
+            return Tier.FREE
+
+        settings = get_settings()
+        # Determine tier based on stripe_price_id
+        if billing.stripe_price_id == settings.stripe_cloud_team_flat_price_id:
```
brenkao, line 48: BillingService has a `get_tier_from_billing`. Could we use that?

if using BillingService is giving circular dependencies, we could remove all the imports from `services/__init__.py`. and directly import from each service.

I'd rather fix the circular dependencies than maintain two copies of calculating tier

PR #601, opened 2025-07-08
Implement lilypad.configure({ auto_llm: true }) for OpenAI in Lilypad TypeScript SDK

koxudaxi:


sdks/typescript/AUTO_LLM_CONSTRAINTS.md
```diff
@@ -0,0 +1,77 @@
...
+```
+This ensures hooks are installed before any application code runs.
+
+### 3. Manual Wrapping (Most Control)
+```typescript
+import OpenAI from "openai";
+import { wrapOpenAI } from "lilypad";
+
+const client = wrapOpenAI(new OpenAI());
+```
```
koxudaxi, line 61: @brenkao @willbakst 
Unlike Python, there appear to be limitations when applying patches (module hooks). If you know of a solution, please share it.
We should also decide and document which approach to recommend as a library before publishing the package. Related to this, I've discovered that things become even more complicated when parent spans exist, so I'll comment on that in a separate PR.

PR #606, opened 2025-07-10
Implement lilypad.trace in Lilypad TypeScript SDK (without versioning)

koxudaxi:


sdks/typescript/src/register.ts
```diff
@@ -18,14 +18,15 @@ import { JSONSpanExporter } from './exporters/json-exporter';
 // Enable debug logging based on environment variable
 const logLevel = process.env.LILYPAD_LOG_LEVEL || 'info';
 if (logLevel && ['debug', 'info', 'warn', 'error'].includes(logLevel)) {
-  logger.setLevel(logLevel as any);
+  logger.setLevel(logLevel as 'debug' | 'info' | 'warn' | 'error');
```
brenkao, line 21: make this into type? I assume we will be using this throughout the typescript sdk

sdks/typescript/src/trace.ts
```diff
@@ -0,0 +1,739 @@
...
+ * Get or create function in Lilypad backend
+ */
+async function getOrCreateFunction(
+  fn: (...args: unknown[]) => unknown,
+  projectId: string,
+  apiKey: string,
+  baseUrl: string,
+): Promise<string | null> {
+  try {
+    const closure = getCachedClosure(fn);
```
brenkao, line 89: I would recommend destructuring this e.g.
```
{ hash, code, name, signature } = getCachedClosure(fn)
```
Then below, you can just do
```
    const createData: FunctionCreate = {
      project_uuid: projectId,
      code,
      hash,
      name,
      signature,
      arg_types: {},
      dependencies,
      is_versioned: false,
      prompt_template: '',
    };

File: sdks/typescript/src/utils/settings.ts [hunk 3]
```diff
@@ -1,5 +1,7 @@
 import type { LilypadConfig } from '../types';
 
+export type LilypadSettings = LilypadConfig;
```
brenkao, line 3: What is this for?

PR #647, opened 2025-08-05
test: Implement 100% test coverage for OpenAI OpenTelemetry support

koxudaxi:


koxudaxi:
* **#655** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/655?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#654** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/654?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#653** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/653?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#652** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/652?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#651** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/651?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#647** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/647?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/647?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#646** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/646?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v1`


<h2></h2>

willbakst:
This PR is way way way too big.

It needs to be broken down. At a glance, it seems like a potential breakdown would be:
1. Structure (i.e. things like `pyproject.toml`, GitHub Actions, etc.)
2. Tests for each individual file to get it to 100% coverage. The PR for (1) above should exclude everything we don't yet test from coverage, and then each PR following can systematically remove exclusions as we add tests.

Another note worth pointing out is that we should have code with tests in the same PR. It's bad practice to separate them. I know we're doing a big PR before this with the OpenAI port, but that was also a monster PR that was difficult to review.

Lastly, @teamdandelion is using `vcr.py` in Mirascope `v2` in order to avoid mocking. We're also using `inline-snapshot`. I think we should consider using those two packages here as well to reduce test complexity.

willbakst:
Some steps for cleanup:
1. Copy the entire TypeScript folder where you can easily access it / paste it.
2. `gt checkout v1` to checkout the `v1` trunk branch.
3. Paste the TypeScript folder back into the `v1` branch.
4. Run `gt add -A` and then `gt create -m "feat: TypeScript OpenAI OpenTelemetry Support"`
5. Run `gt get "LILYPAD-227"` to checkout this branch.
6. Delete / remove the TypeScript changes and run `gt modify -a` + `gt submit` to clean up this PR so it's only Python.

This will create a new stack from `v1` with the TypeScript changes.

Next, do the same thing for the upstream testing stack. Take the TypeScript changes, apply them to the branch for TypeScript created above on top of `v1`, and create a second item in that stack with those changes. Then return to the Python testing stack and remove the TypeScript changes to clean up the PR.

Once you cleanly split this into two stacks (one for Python, one for TypeScript), you should then further split the Python testing stack into smaller, logical units as described above (namely CI followed by per-file tests). The CI PR should exclude all files temporarily since there will be no tests (so that coverage still passes with CI running). Then, in every follow-up PR, include a small change to the tests (e.g. one file) and update the coverage configuration so that coverage runs on that file (and any other files down stack that have already had their tests added).

The end result here should be a Python stack and a TypeScript stack. In the Python stack, you should have the downstack OTel implementation changes (which has comments you should address), followed by this PR which will be split into the CI and single file test change PRs.

Once the entire stack is reviewed and merged, we should have a full OpenAI OTel instrumentation implementation with 100% coverage.

.github/workflows/python-sdk-ci.yml
willbakst, line 1: I'm wondering if it would be better to have a single `lint.yml` and `test.yml` that tests each piece so that are linting is colocated, all testing is colocated, etc.? I believe we can still filter jobs to only run when changes are made to specific paths using something like [this](https://github.com/dorny/paths-filter).

Each job would then be something like `lint-python-sdk` with a name like `Lint Python SDK`.

One reason for wanting this would be that we could ensure at the top level that running `bun run test:ci` is equivalent to running the `test.yml` CI. Same with `bun run lint:ci`.

sdks/python/.pre-commit-config.yaml
willbakst, line 1: I would prefer to use the top level husky pre-commit stuff rather than have separate pre-commit handling for each piece. We already have it at the top level for the monorepo, so we may as well use it. When developing on this, we'll be developing at the top level anyway and not from directly inside the SDK package.

We can update stuff like `bun run lint:ci` to include this, and we can add specific scripts like `bun run lint:python-sdk`?

In line with this, we should add top level scripts for all of this so that it's easy to run tests / lint / etc. from the top level.

The [Mirascope `v2` package.json](https://github.com/Mirascope/mirascope/blob/v2/package.json) is a good example that we should mimic here.

sdks/python/pyproject.toml
```diff
@@ -149,25 +158,46 @@ known-first-party = ["lilypad", "tests"]
 pythonVersion = "3.10"
 venvPath = "."
 venv = ".venv"
+include = ["src", "tests"]
+exclude = [".venv", "examples"]
 
-exclude = [".venv", "tests", "src/lilypad", "examples"]
+[tool.pytest.ini_options]
+testpaths = ["tests"]
```
willbakst, line 164: are these not the defaults?

sdks/python/pyproject.toml
```diff
@@ -149,25 +158,46 @@ known-first-party = ["lilypad", "tests"]
...
+    "--cov-report=xml",
+#    "--cov-fail-under=100",
+]
+asyncio_mode = "auto"
 
 [tool.coverage.run]
 branch = true
 source = ["src/lilypad"]
+omit = [
+    "*/tests/*",
```
willbakst: why are we excluding tests from coverage?

sdks/python/pyproject.toml
```diff
@@ -149,25 +158,46 @@ known-first-party = ["lilypad", "tests"]
...
+python_files = ["test_*.py"]
+python_classes = ["Test*"]
+python_functions = ["test_*"]
+addopts = [
+    "-v",
+    "--cov=lilypad",
+    "--cov-report=term-missing",
+    "--cov-report=html",
+    "--cov-report=xml",
+#    "--cov-fail-under=100",
```
willbakst: remove this if unneeded?

sdks/python/pyproject.toml
```diff
@@ -149,25 +158,46 @@ known-first-party = ["lilypad", "tests"]
 pythonVersion = "3.10"
 venvPath = "."
 venv = ".venv"
+include = ["src", "tests"]
+exclude = [".venv", "examples"]
```
willbakst: why aren't we running pyright on exapmles? also shouldn't it already know not to run on `.venv` given we mention it explicitly two lines above?

.github/workflows/python-sdk-ci.yml
```diff
@@ -0,0 +1,91 @@
+name: Python SDK CI
+
+on:
+  push:
+    branches: [main]
```
willbakst: FYI this should be `[v1]` for now since we're only operating on the `v1` branch and not `main` yet.

sdks/python/pyproject.toml
koxudaxi: I should remove it.

.github/workflows/lint.yml
```diff
@@ -33,3 +33,63 @@ jobs:
 
       - name: Run lint
         run: bun run lint
+
+  lint-python-sdk:
+    name: Lint Python SDK
+    runs-on: ubuntu-latest
+    if: |
```
willbakst: This feels brittle. What happens if I don't include `python` or `sdk` in my PR but I make changes to the Python SDK?

What about something more like this?

```yml
jobs:
  # Detect which paths changed
  changes:
    runs-on: ubuntu-latest
    outputs:
      python-sdk: ${{ steps.changes.outputs.python-sdk }}
    steps:
      - uses: actions/checkout@v4
      - uses: dorny/paths-filter@v2
        id: changes
        with:
          filters: |
            python-sdk:
              - 'sdks/python/**'
              - '.github/workflows/test.yml'

  # Python SDK tests
  lint-python-sdk:
    needs: changes
    if: ${{ needs.changes.outputs.python-sdk == 'true' }}
    runs-on: ubuntu-latest
    ...
```

.github/workflows/test.yml
```diff
@@ -41,3 +41,70 @@ jobs:
 
       - name: Run tests
         run: bun run test:coverage
+
+  test-python-sdk:
+    name: Test Python SDK (Python ${{ matrix.python-version }})
+    runs-on: ${{ matrix.os }}
+    if: |
```
willbakst: same thing here. see earlier comment

sdks/python/src/lilypad/_internal/otel/openai/utils.py
```diff
@@ -139,10 +140,10 @@ def default_openai_cleanup(
 
 
 def get_tool_calls(
-    message: dict | "ChatCompletionMessage",
-) -> list[dict[str, Any]] | None:
+    message: Union[dict, "ChatCompletionMessage"],
```
willbakst: We only support 3.10+ so we should use the `|` operator, not `Union`.

sdks/python/src/lilypad/_internal/otel/openai/utils.py
```diff
@@ -153,12 +154,8 @@ def get_tool_calls(
 
     calls = []
     for tool_call in tool_calls:
-        tool_call_dict = {}
-        if hasattr(tool_call, "id"):
-            call_id = tool_call.id
-            tool_type = tool_call.type
-            tool_call_dict["function"] = tool_call.function.model_dump(mode="python")
-        else:
+        tool_call_dict: dict[str, Any] = {}
```
willbakst: in line with comments down stack, we should avoid using `Any` unless absolutely necessary

.github/workflows/test.yml
```diff
@@ -7,6 +7,21 @@ on:
...
+      - uses: actions/checkout@v4
+      - uses: dorny/paths-filter@v2
+        id: changes
+        with:
+          filters: |
+            python-sdk:
+              - 'sdks/python/**'
+              - '.github/workflows/test.yml'
+
   test:
```
willbakst: nit: we should rename this to `test-app` and do the same filtering right?

.github/workflows/lint.yml
```diff
@@ -10,6 +10,21 @@ on:
...
+        id: changes
+        with:
+          filters: |
+            python-sdk:
+              - 'sdks/python/**'
+              - '.github/workflows/lint.yml'
+
   codespell:
     name: Spell Check
     runs-on: ubuntu-latest
```
willbakst, line 34: <p class='graphite__hidden'><i>[Re: line +36]</i></p>
nit: we should rename this to `lint-app` and do the same filtering, right?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/647?utm_source=unchanged-line-comment">Graphite</a>.</p>

willbakst (approved):

willbakst:
### Merge activity

* **Aug 12, 12:46 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/lilypad/647).

PR #652, opened 2025-08-06
test: add OpenTelemetry utility tests for buffers and server address handling

koxudaxi:
### TL;DR

Added unit tests for OpenTelemetry utility functions in the Lilypad SDK.

### What changed?

- Created new test files for the OpenTelemetry utilities in the Lilypad SDK
- Added comprehensive tests for `ChoiceBuffer` class, including initialization and methods for appending text content and tool calls
- Added tests for `ToolCallBuffer` class, covering initialization and argument appending
- Added tests for the `set_server_address_and_port` function, verifying it correctly extracts server information from different URL formats

### How to test?

Run the Python tests for the Lilypad SDK:

```bash
cd sdks/python
pytest tests/lilypad/_internal/otel/test_otel_utils.py -v
```

### Why make this change?

These tests ensure the OpenTelemetry utility functions work correctly, particularly the buffer classes that handle streaming responses and the server address extraction function. This improves reliability of telemetry data collection in the Lilypad SDK and helps prevent regressions in future changes.

PR #651, opened 2025-08-06
test: add tests for Lilypad configuration and utils

koxudaxi:
### TL;DR

Added unit tests for the Lilypad SDK's internal configuration and utilities modules.

### What changed?

- Created test directory structure for the Lilypad Python SDK
- Added comprehensive tests for the `CryptoIdGenerator` class in the configuration module
- Added tests for the `configure()` function with various parameters and scenarios
- Added tests for the `json_dumps()` utility function with different data types and structures

### How to test?

Run the tests using pytest:

```bash
cd sdks/python
pytest tests/lilypad/_internal/test_configuration.py tests/lilypad/_internal/test_utils.py -v
```

### Why make this change?

These tests improve code coverage and ensure the reliability of the Lilypad SDK's internal components. The tests verify:

1. Proper generation of unique span and trace IDs
2. Correct handling of invalid IDs
3. Proper configuration of logging and tracing
4. Correct JSON serialization of various data types including dataclasses, UUIDs, and special characters

This helps maintain code quality and prevents regressions as the SDK evolves.

PR #653, opened 2025-08-06
test: add OpenAI instrumentation tests

koxudaxi:
### TL;DR

Added unit tests for the OpenAI instrumentation module in the Lilypad SDK.

### What changed?

Added comprehensive unit tests for the `instrument_openai` function that handles OpenAI client instrumentation for telemetry. The tests cover:

- Instrumenting synchronous OpenAI clients
- Instrumenting asynchronous OpenAI clients
- Handling package version not found errors
- Handling wrapper failures for both sync and async clients
- Verifying debug logging functionality

### How to test?

Run the unit tests with pytest:

```bash
cd sdks/python
pytest tests/lilypad/_internal/otel/openai/test_instrument.py -v
```

### Why make this change?

These tests ensure the OpenAI instrumentation module works correctly across different scenarios and client types. They verify that the telemetry functionality properly wraps OpenAI client methods and handles error conditions gracefully, improving the reliability of the OpenTelemetry integration.

### DIFF
--- a/sdks/python/src/lilypad/_internal/otel/openai/instrument.py
+++ b/sdks/python/src/lilypad/_internal/otel/openai/instrument.py
@@ -21,15 +21,14 @@ create OpenTelemetry spans for API calls.
 # Modifications copyright (C) 2025 Mirascope
 
 import logging
-from importlib.metadata import version, PackageNotFoundError
+from importlib.metadata import PackageNotFoundError, version
 
-from wrapt import wrap_function_wrapper, FunctionWrapper
+from wrapt import FunctionWrapper, wrap_function_wrapper
+from openai import OpenAI, AsyncOpenAI
 from opentelemetry.trace import get_tracer
 from opentelemetry.semconv.schemas import Schemas
 
 from . import _patch
-
-from openai import OpenAI, AsyncOpenAI
 
 logger = logging.getLogger(__name__)
 
@@ -41,6 +40,10 @@ def instrument_openai(client: OpenAI | AsyncOpenAI) -> None:
     Args:
         client: The OpenAI client instance to instrument.
     """
+    if hasattr(client, "_lilypad_instrumented"):
+        logger.debug(f"{type(client).__name__} already instrumented, skipping")
+        return
+
     try:
         lilypad_version = version("lilypad-sdk")
     except PackageNotFoundError:
@@ -101,3 +104,5 @@ def instrument_openai(client: OpenAI | AsyncOpenAI) -> None:
             logger.debug("Successfully wrapped Completions.parse")
         except Exception as e:
             logger.warning(f"Failed to wrap Completions.parse: {type(e).__name__}: {e}")
+
+    client._lilypad_instrumented = True  # pyright: ignore[reportAttributeAccessIssue]
--- a/sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
+++ b/sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
@@ -1,0 +1,138 @@
+"""Tests for OpenAI instrumentation."""
+
+import inspect
+
+from openai import OpenAI, AsyncOpenAI
+
+from lilypad._internal.otel.openai.instrument import instrument_openai
+
+
+def test_sync_client_methods_are_patched():
+    """Test that OpenAI sync client methods are patched while maintaining signatures."""
+    client = OpenAI(api_key="test")
+    original_create = client.chat.completions.create
+    original_parse = client.chat.completions.parse
+
+    original_create_sig = inspect.signature(original_create)
+    original_parse_sig = inspect.signature(original_parse)
+
+    instrument_openai(client)
+
+    assert client.chat.completions.create is not original_create
+    assert client.chat.completions.parse is not original_parse
+
+    assert inspect.signature(client.chat.completions.create) == original_create_sig
+    assert inspect.signature(client.chat.completions.parse) == original_parse_sig
+
+
+def test_async_client_methods_are_patched():
+    """Test that AsyncOpenAI client methods are patched while maintaining signatures."""
+    client = AsyncOpenAI(api_key="test")
+    original_create = client.chat.completions.create
+    original_parse = client.chat.completions.parse
+
+    original_create_sig = inspect.signature(original_create)
+    original_parse_sig = inspect.signature(original_parse)
+
+    instrument_openai(client)
+
+    assert client.chat.completions.create is not original_create
+    assert client.chat.completions.parse is not original_parse
+
+    assert inspect.signature(client.chat.completions.create) == original_create_sig
+    assert inspect.signature(client.chat.completions.parse) == original_parse_sig
+
+
+def test_patched_methods_are_callable():
+    """Test that patched methods remain callable."""
+    client = OpenAI(api_key="test")
+    instrument_openai(client)
+
+    assert callable(client.chat.completions.create)
+    assert callable(client.chat.completions.parse)
+
+
+def test_patched_methods_have_wrapped_attribute():
+    """Test that patched methods have __wrapped__ attribute for introspection."""
+    client = OpenAI(api_key="test")
+
+    instrument_openai(client)
+
+    assert hasattr(client.chat.completions.create, "__wrapped__")
+    assert hasattr(client.chat.completions.parse, "__wrapped__")
+    assert callable(client.chat.completions.create.__wrapped__)  # pyright: ignore[reportAttributeAccessIssue]
+    assert callable(client.chat.completions.parse.__wrapped__)  # pyright: ignore[reportAttributeAccessIssue]
+
+
+def test_client_marked_as_instrumented():
+    """Test that client is marked as instrumented after instrumentation."""
+    client = OpenAI(api_key="test")
+
+    assert not hasattr(client, "_lilypad_instrumented")
+
+    instrument_openai(client)
+
+    assert hasattr(client, "_lilypad_instrumented")
+    assert client._lilypad_instrumented is True  # pyright: ignore[reportAttributeAccessIssue]
+
+
+def test_instrumentation_idempotent():
+    """Test that instrumentation is truly idempotent."""
+    client = OpenAI(api_key="test")
+
+    instrument_openai(client)
+    first_create = client.chat.completions.create
+    first_parse = client.chat.completions.parse
+
+    instrument_openai(client)
+    second_create = client.chat.completions.create
+    second_parse = client.chat.completions.parse
+
+    assert second_create is first_create
+    assert second_parse is first_parse
+
+
+def test_async_instrumentation_idempotent():
+    """Test that async instrumentation is truly idempotent."""
+    client = AsyncOpenAI(api_key="test")
+
+    instrument_openai(client)
+    first_create = client.chat.completions.create
+    first_parse = client.chat.completions.parse
+
+    instrument_openai(client)
+    second_create = client.chat.completions.create
+    second_parse = client.chat.completions.parse
+
+    assert second_create is first_create
+    assert second_parse is first_parse
+
+
+def test_preserves_async_nature():
+    """Test that async methods remain async after patching."""
+    from pydantic import BaseModel
+
+    class TestResponse(BaseModel):
+        test: str
+
+    client = AsyncOpenAI(api_key="test")
+
+    create_result_before = client.chat.completions.create(model="test", messages=[])
+    parse_result_before = client.chat.completions.parse(
+        model="test", messages=[], response_format=TestResponse
+    )
+    assert inspect.iscoroutine(create_result_before)
+    assert inspect.iscoroutine(parse_result_before)
+    create_result_before.close()
+    parse_result_before.close()
+
+    instrument_openai(client)
+
+    create_result_after = client.chat.completions.create(model="test", messages=[])
+    parse_result_after = client.chat.completions.parse(
+        model="test", messages=[], response_format=TestResponse
+    )
+    assert inspect.iscoroutine(create_result_after)
+    assert inspect.iscoroutine(parse_result_after)
+    create_result_after.close()
+    parse_result_after.close()


koxudaxi:
* **#655** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/655?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#654** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/654?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#653** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/653?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/653?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#652** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/652?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#651** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/651?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#647** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/647?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#646** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/646?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v1`


<h2></h2>

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
willbakst, line 1: I'm wondering if all of this mocking is worthwhile. What do we gain from these types of tests?

If we did this:
```python
from openai import OpenAI

def test_instrument_openai():
    client = OpenAI()
    instrument_openai(client)
    # TESTS HERE
```

What would we want for the `# TESTS HERE` section to test that `instrument_openai` is doing what we want? Would it be sufficient to ensure that the desired methods are in fact patched correctly?

My question really boils down to what the impact of all of this mocking will have as we potentially update underlying implementation details without actually impacting the interface itself. If calling `instrument_openai` has the same behavior from the consumers perspective, changing the underlying implementation shouldn't break our tests.

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
```diff
@@ -0,0 +1,135 @@
+"""Tests for OpenAI instrumentation."""
+
+from lilypad._internal.otel.openai.instrument import instrument_openai
+
+
+def test_sync_client_methods_are_patched():
+    """Test that OpenAI sync client methods are patched."""
+    from openai import OpenAI
```
willbakst: nit: we can just pull these out into single imports at the top level

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
```diff
@@ -0,0 +1,135 @@
...
+
+    client = OpenAI(api_key="test")
+    instrument_openai(client)
+
+    assert hasattr(client.chat.completions.create, "__wrapped__")
+    assert hasattr(client.chat.completions.parse, "__wrapped__")
+
+
+def test_patched_methods_callable_with_args():
```
willbakst: this test is no different from the above one just checking callability?

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
```diff
@@ -0,0 +1,135 @@
...
+def test_instrumentation_idempotent():
+    """Test that instrumentation can be applied multiple times without error."""
+    from openai import OpenAI
+
+    client = OpenAI(api_key="test")
+
+    for _ in range(3):
+        instrument_openai(client)
+
+    assert callable(client.chat.completions.create)
```
willbakst: not sure how ensuring it's callable ensures idempotency?

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
```diff
@@ -0,0 +1,135 @@
...
+    """Test that OpenAI sync client methods are patched."""
+    from openai import OpenAI
+
+    client = OpenAI(api_key="test")
+    original_create = client.chat.completions.create
+    original_parse = client.chat.completions.parse
+
+    instrument_openai(client)
+
+    assert client.chat.completions.create is not original_create
```
willbakst, line 21: Can we expand on this test to ensure that while different they are exactly the same? Something like matching signatures? The idea of the test being that we've done something to the function (so it's not the same) but we can reasonably expect it to operate exactly as we want?

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
```diff
@@ -0,0 +1,135 @@
...
+    original_create = client.chat.completions.create
+
+    instrument_openai(client)
+    first_create = client.chat.completions.create
+
+    instrument_openai(client)
+    second_create = client.chat.completions.create
+
+    assert first_create is not original_create
+    assert second_create is not first_create
```
willbakst: Hmm, wouldn't we want `second_create is first_create`? Shouldn't we recognize we already patched it and thus not do anything?

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
```diff
@@ -0,0 +1,135 @@
...
+
+def test_preserves_async_nature():
+    """Test that async methods remain async after patching."""
+    from openai import AsyncOpenAI
+    import inspect
+
+    client = AsyncOpenAI(api_key="test")
+    instrument_openai(client)
+
+    assert inspect.iscoroutinefunction(client.chat.completions.create) or callable(
```
willbakst: Hmm, this `or` clause doesn't make sense to me. This will still pass if it's a sync function, which defeats the purpose of the test?

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
```diff
@@ -0,0 +1,135 @@
...
+        client.chat.completions.create
+    )
+
+
+def test_different_client_types():
```
willbakst: We already test these above in the sync vs. async cases? This test seems redundant.

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
```diff
@@ -0,0 +1,135 @@
...
+    async_client = AsyncOpenAI(api_key="test")
+
+    instrument_openai(sync_client)
+    instrument_openai(async_client)
+
+    assert hasattr(sync_client.chat.completions.create, "__wrapped__")
+    assert hasattr(async_client.chat.completions.create, "__wrapped__")
+
+
+def test_both_create_and_parse_are_patched():
```
willbakst: Why not just include this as part of the first patching test? Why separate it?

sdks/python/tests/lilypad/_internal/test_configuration.py
```diff
@@ -4,7 +4,7 @@
 from opentelemetry import trace
 from opentelemetry.trace import INVALID_SPAN_ID, INVALID_TRACE_ID
 
-from lilypad.configuration import CryptoIdGenerator, configure
+from lilypad._internal.configuration import CryptoIdGenerator, configure
```
willbakst: Down stack `configuration` was made public outside of `_internal`, right? What is this change?

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
koxudaxi: This is to ensure idempotency.

sdks/python/tests/lilypad/_internal/test_configuration.py
koxudaxi: I made mistake. I will revert it.

sdks/python/src/lilypad/_internal/configuration.py
willbakst, line 1: What's going on here? Don't we have `configuration.py` at the same level as `_internal` and not inside of `_internal`? Is this a mistaken duplicate?

sdks/python/src/lilypad/__init__.py
```diff
@@ -1,7 +1,7 @@
 """Lilypad SDK - The official Python library for the Lilypad API."""
 
-from . import configuration
-from .configuration import configure
+from ._internal import configuration
```
willbakst: look like this is somehow incorrectly reverting changes from down stack?

sdks/python/src/lilypad/configuration.py
```diff
@@ -1,60 +1,5 @@
-"""Configuration utilities for Lilypad SDK initialization and setup."""
+"""Configuration module for Lilypad SDK."""
 
-import logging
-from secrets import token_bytes
-from opentelemetry import trace
-from opentelemetry.trace import INVALID_SPAN_ID, INVALID_TRACE_ID
-from opentelemetry.sdk.trace.id_generator import IdGenerator
-from rich.logging import RichHandler
+from ._internal.configuration import CryptoIdGenerator, configure
```
willbakst: If you're doing this change on purpose, I don't understand it. Can you explain why you're making this change?

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
```diff
@@ -0,0 +1,140 @@
...
+    assert hasattr(client.chat.completions.create, "__wrapped__")
+    assert hasattr(client.chat.completions.parse, "__wrapped__")
+
+
+def test_patched_methods_are_callable():
+    """Test that patched methods remain callable."""
+    client = OpenAI(api_key="test")
+    instrument_openai(client)
+
+    assert callable(client.chat.completions.create)
```
willbakst, line 51: we already test this above, either remove these assertions from the above tests or remove this test

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
```diff
@@ -0,0 +1,140 @@
...
+    assert callable(client.chat.completions.parse)
+
+
+def test_patched_methods_have_wrapped_attribute():
+    """Test that patched methods have __wrapped__ attribute for accessing original."""
+    client = OpenAI(api_key="test")
+    instrument_openai(client)
+
+    assert hasattr(client.chat.completions.create, "__wrapped__")
```
willbakst, line 61: same thing here

I think my preference would be to keep these individual tests and remove these specific assertions from the test above, which should be focused more on signature parity

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
```diff
@@ -0,0 +1,140 @@
...
+def test_instrumentation_idempotent():
+    """Test that instrumentation can be applied multiple times without error."""
+    client = OpenAI(api_key="test")
+
+    original_create = client.chat.completions.create
+    original_parse = client.chat.completions.parse
+    original_sig_create = inspect.signature(original_create)
+    original_sig_parse = inspect.signature(original_parse)
+
+    for _i in range(3):
```
willbakst: Hmm, shouldn't we detect in the code if `instrument_openai` has already been run and do nothing? In which case, shouldn't we then test here that after we call it once, the `create` and `parse` methods will actually be truly `==` equal on any subsequent instrumentation attempts?

willbakst: what's with the underscore `_i`? Why not just manually run instrumentation once, then pull the methods, then run it again, and then check that the methods are equal in line with my other comment?

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
willbakst: But we ensure idempotency in the above test about idempotency?

sdks/python/tests/lilypad/_internal/test_configuration.py
```diff
@@ -37,7 +38,7 @@ def test_generate_trace_id() -> None:
     assert all(0 < tid < 2**128 for tid in trace_ids)
 
 
-@patch("lilypad.configuration.token_bytes")
+@patch("lilypad._internal.configuration.token_bytes")
```
willbakst: nit: in line with other comment, this shouldn't be `_internal`?

sdks/python/src/lilypad/configuration.py
koxudaxi: It seems that I misunderstood the previous comment and made a change. I will correct it.

sdks/python/src/lilypad/_internal/otel/openai/instrument.py
```diff
@@ -41,6 +41,10 @@ def instrument_openai(client: OpenAI | AsyncOpenAI) -> None:
     Args:
         client: The OpenAI client instance to instrument.
     """
+    if hasattr(client, "_lilypad_instrumented"):
+        logger.debug("Client already instrumented, skipping")
```
willbakst: nit: we should make this more specific than just `Client` right? Shouldn't we make it specific to OpenAI? Or specific to the client they are attempting to instrument?

sdks/python/tests/lilypad/_internal/otel/openai/test_instrument.py
```diff
@@ -0,0 +1,145 @@
...
+    first_parse = client.chat.completions.parse
+
+    instrument_openai(client)
+    second_create = client.chat.completions.create
+    second_parse = client.chat.completions.parse
+
+    assert second_create is first_create
+    assert second_parse is first_parse
+
+    instrument_openai(client)
```
willbakst: nit: why does this have three and the below test have two? I think just two is enough, so I opt for removing the third round of instrumentation in this test

willbakst (approved):

willbakst:
### Merge activity

* **Aug 12, 12:53 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/lilypad/653).

PR #670, opened 2025-08-14
Port Google OpenTelemetry into Lilypad v1

koxudaxi:
### TL;DR

Add Google AI (Gemini) instrumentation support to the Lilypad SDK, allowing users to collect telemetry data from Google's AI models.

### What changed?

- Added Google AI instrumentation capabilities to the Lilypad Python SDK
- Implemented OpenTelemetry tracing for Google's `generate_content` and streaming methods
- Created utility functions to extract and format Google AI response data for telemetry
- Added a new example demonstrating how to use the Google instrumentation
- Updated dependencies to include `google-genai`, `pillow`, and `proto-plus` packages

### How to test?

1. Install the SDK with Google support:
```bash
pip install "lilypad[google]"
```

2. Run the new example:
```bash
python sdks/python/examples/instrument_google_with_console.py
```

This example demonstrates how to instrument Google's Gemini models and view telemetry data in the console.

### Why make this change?

This change extends Lilypad's instrumentation capabilities to include Google AI models alongside the existing OpenAI and Anthropic support. This allows users to collect consistent telemetry data across multiple LLM providers, enabling better monitoring, debugging, and analysis of AI model interactions.

### DIFF
--- a/sdks/python/src/lilypad/_internal/otel/google/_patch.py
+++ b/sdks/python/src/lilypad/_internal/otel/google/_patch.py
@@ -1,0 +1,111 @@
+"""Patching utilities for Google client methods to add OpenTelemetry instrumentation.
+
+This module contains the core logic for wrapping Google AI calls with telemetry spans.
+"""
+
+from __future__ import annotations
+
+import json
+from typing import Any, cast
+
+from opentelemetry.trace import Span, Tracer
+from opentelemetry.util.types import AttributeValue
+from typing_extensions import ParamSpec
+
+from ..base.protocols import (
+    AsyncCompletionHandler,
+    SyncCompletionHandler,
+)
+from ..base.wrappers import (
+    create_async_wrapper,
+    create_sync_wrapper,
+)
+from . import _utils
+from google.genai.types import (
+    GenerateContentConfig,
+    GenerateContentConfigDict,
+    ContentListUnionDict,
+    ContentUnionDict,
+)
+from google.genai.models import AsyncModels, Models
+from google.genai.types import GenerateContentResponse
+
+P = ParamSpec("P")
+
+
+def _process_messages(
+    span: Span,
+    kwargs: dict[str, Any],
+) -> None:
+    """Process and record input messages."""
+    config = cast(
+        GenerateContentConfig | GenerateContentConfigDict | None, kwargs.get("config")
+    )
+    if config:
+        if isinstance(config, dict):
+            system_instruction = config.get("system_instruction")
+        else:
+            system_instruction = config.system_instruction
+
+        if system_instruction:
+            span.add_event(
+                "gen_ai.system.message",
+                attributes={
+                    "gen_ai.system": "google_genai",
+                    "gen_ai.content": str(system_instruction),
+                },
+            )
+
+    contents = cast(ContentListUnionDict | None, kwargs.get("contents"))
+    if not contents:
+        return
+
+    if not isinstance(contents, list):
+        contents = [contents]
+
+    messages = _utils._convert_content_to_messages(
+        cast(list[ContentUnionDict], contents)
+    )
+
+    for message in messages:
+        role = message.get("role", "user")
+        event_name = f"gen_ai.{role}.message"
+        attributes: dict[str, AttributeValue] = {
+            "gen_ai.system": "google_genai",
+        }
+
+        content = message.get("content")
+        if content:
+            attributes["gen_ai.content"] = content
+
+        tool_calls = message.get("tool_calls")
+        if tool_calls:
+            attributes["gen_ai.tool_calls"] = json.dumps(tool_calls)
+
+        span.add_event(event_name, attributes=attributes)
+
+
+def generate_content_patch_factory(
+    tracer: Tracer,
+) -> SyncCompletionHandler[P, GenerateContentResponse, Models]:
+    return create_sync_wrapper(
+        tracer=tracer,
+        get_span_attributes=_utils.get_llm_request_attributes,
+        process_messages=_process_messages,
+        create_stream_wrapper=None,
+        process_response=_utils.process_response,
+        handle_stream=False,
+    )
+
+
+def async_generate_content_patch_factory(
+    tracer: Tracer,
+) -> AsyncCompletionHandler[P, GenerateContentResponse, AsyncModels]:
+    return create_async_wrapper(
+        tracer=tracer,
+        get_span_attributes=_utils.get_llm_request_attributes,
+        process_messages=_process_messages,
+        process_response=_utils.process_response,
+        create_async_stream_wrapper=None,
+        handle_stream=False,
+    )
--- a/sdks/python/src/lilypad/_internal/otel/google/_utils.py
+++ b/sdks/python/src/lilypad/_internal/otel/google/_utils.py
@@ -1,0 +1,313 @@
+"""Utility functions for Google OpenTelemetry instrumentation.
+
+This module provides helper functions for extracting and formatting Google AI
+response data for telemetry purposes.
+"""
+
+from __future__ import annotations
+
+import base64
+import json
+from enum import Enum
+from io import BytesIO
+from typing import Any, cast
+
+import PIL
+import PIL.WebPImagePlugin
+from opentelemetry.trace import Span
+from opentelemetry.util.types import AttributeValue
+from opentelemetry.semconv._incubating.attributes import gen_ai_attributes
+from google.genai.types import (
+    GenerateContentConfig,
+    GenerateContentConfigDict,
+    FinishReason,
+    ContentUnionDict,
+    ContentDict,
+    Part,
+    PartDict,
+)
+from ..utils import (
+    BaseMetadata,
+    set_server_address_and_port,
+)
+from ..types import (
+    LLMOpenTelemetryFunctionCall,
+    LLMOpenTelemetryMessage,
+    LLMOpenTelemetryToolCall,
+    LLMOpenTelemetryInlineData,
+)
+from google.genai.types import GenerateContentResponse, Content
+
+
+class GoogleMetadata(BaseMetadata, total=False):
+    model_version: str | None
+    response_id: str | None
+    safety_ratings: list[dict[str, Any]] | None
+    citation_metadata: dict[str, Any] | None
+    input_tokens: int
+    output_tokens: int
+
+
+def _extract_tool_names_from_config(
+    config: GenerateContentConfigDict,
+) -> list[str] | None:
+    if not (tools := config.get("tools")):
+        return None
+
+    function_names: list[str] = []
+
+    for tool in tools:
+        if isinstance(tool, dict) and (
+            function_declarations := tool.get("function_declarations")
+        ):
+            for declaration in function_declarations:
+                if name := declaration.get("name"):
+                    function_names.append(name)
+    return function_names
+
+
+def get_llm_request_attributes(
+    kwargs: dict[str, Any],
+    client: Any | None,
+    operation_name: str = gen_ai_attributes.GenAiOperationNameValues.CHAT.value,
+) -> dict[str, AttributeValue]:
+    attributes: dict[str, AttributeValue] = {
+        gen_ai_attributes.GEN_AI_SYSTEM: "google_genai",
+        gen_ai_attributes.GEN_AI_REQUEST_MODEL: str(kwargs.get("model", "unknown")),
+        gen_ai_attributes.GEN_AI_OPERATION_NAME: operation_name,
+    }
+
+    if client:
+        set_server_address_and_port(client, attributes)
+
+    config = cast(
+        GenerateContentConfig | GenerateContentConfigDict | None, kwargs.get("config")
+    )
+    if config:
+        if isinstance(config, GenerateContentConfig):
+            config = cast(GenerateContentConfigDict, config.to_json_dict())
+        if (temperature := config.get("temperature")) and temperature is not None:
+            attributes[gen_ai_attributes.GEN_AI_REQUEST_TEMPERATURE] = temperature
+
+        if (top_p := config.get("top_p")) and top_p is not None:
+            attributes[gen_ai_attributes.GEN_AI_REQUEST_TOP_P] = top_p
+        if (top_k := config.get("top_k")) and top_k is not None:
+            attributes[gen_ai_attributes.GEN_AI_REQUEST_TOP_K] = top_k
+        if max_output_tokens := config.get("max_output_tokens"):
+            attributes[gen_ai_attributes.GEN_AI_REQUEST_MAX_TOKENS] = max_output_tokens
+
+        if stop_sequences := config.get("stop_sequences"):
+            attributes[gen_ai_attributes.GEN_AI_REQUEST_STOP_SEQUENCES] = stop_sequences
+        if "presence_penalty" in config and config["presence_penalty"] is not None:
+            attributes[gen_ai_attributes.GEN_AI_REQUEST_PRESENCE_PENALTY] = config[
+                "presence_penalty"
+            ]
+        if "frequency_penalty" in config and config["frequency_penalty"] is not None:
+            attributes[gen_ai_attributes.GEN_AI_REQUEST_FREQUENCY_PENALTY] = config[
+                "frequency_penalty"
+            ]
+
+        if function_names := _extract_tool_names_from_config(config):
+            attributes["gen_ai.request.functions"] = tuple(function_names)
+
+    return attributes
+
+
+def set_response_attributes(
+    span: Span,
+    response: GenerateContentResponse,
+    metadata: GoogleMetadata | None = None,
+) -> None:
+    attributes: dict[str, AttributeValue] = {}
+
+    if response.model_version:
+        attributes[gen_ai_attributes.GEN_AI_RESPONSE_MODEL] = response.model_version
+
+    if response.response_id:
+        attributes[gen_ai_attributes.GEN_AI_RESPONSE_ID] = response.response_id
+
+    if response.usage_metadata:
+        usage = response.usage_metadata
+        if usage.prompt_token_count:
+            attributes[gen_ai_attributes.GEN_AI_USAGE_INPUT_TOKENS] = (
+                usage.prompt_token_count
+            )
+        if usage.candidates_token_count:
+            attributes[gen_ai_attributes.GEN_AI_USAGE_OUTPUT_TOKENS] = (
+                usage.candidates_token_count
+            )
+
+    if response.candidates:
+        finish_reasons = []
+        for candidate in response.candidates:
+            if isinstance(candidate.finish_reason, Enum):
+                finish_reasons.append(candidate.finish_reason.name)
+
+        if finish_reasons:
+            attributes[gen_ai_attributes.GEN_AI_RESPONSE_FINISH_REASONS] = tuple(
+                finish_reasons
+            )
+
+    span.set_attributes(attributes)
+
+
+def _convert_content_to_messages(
+    contents: list[ContentUnionDict],
+) -> list[LLMOpenTelemetryMessage]:
+    messages: list[LLMOpenTelemetryMessage] = []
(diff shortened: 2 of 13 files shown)


koxudaxi:
* **#671** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/671?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#672](https://github.com/Mirascope/lilypad/pull/672) <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/672?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#673](https://github.com/Mirascope/lilypad/pull/673) <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/673?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#670** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/670?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/670?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v1`


<h2></h2>

sdks/python/src/lilypad/_internal/otel/google/_patch.py
```diff
@@ -0,0 +1,199 @@
...
+
+P = ParamSpec("P")
+
+
+def _process_messages(
+    span: Span,
+    kwargs: dict[str, Any],
+) -> None:
+    """Process and record input messages."""
+    config = cast(
```
willbakst, line 41: This `cast` and the one below are unsafe since `config` can have type `None`. Here is a re-write of this whole block in a way that makes more sense to me:

```python
def _process_messages(
    span: Span,
    kwargs: dict[str, Any],
) -> None:
    """Process and record input messages."""
    config = cast(
        GenerateContentConfig | GenerateContentConfigDict | None, kwargs.get("config")
    )
    if config:
        if isinstance(config, GenerateContentConfig):
            system_instruction = config.system_instruction
        else:
            system_instruction = config.get("system_instruction")

        if system_instruction:
            span.add_event(
                "gen_ai.system.message",
                attributes={
                    "gen_ai.system": "google_genai",
                    "gen_ai.content": str(system_instruction),
                },
            )

    contents = cast(ContentListUnionDict | None, kwargs.get("contents"))
    if not contents:
        return
    
    if not isinstance(contents, list):
        contents = [contents]
    contents = cast(list[ContentUnionDict], [contents])
    
    messages = _process_content_to_messages(contents)

    for message in messages:
        role = message.get("role", "user")
        event_name = f"gen_ai.{role}.message"
        attributes: dict[str, AttributeValue] = {
            "gen_ai.system": "google_genai",
        }

        content = message.get("content")
        if content:
            attributes["gen_ai.content"] = content

        tool_calls = message.get("tool_calls")
        if tool_calls:
            attributes["gen_ai.tool_calls"] = json.dumps(tool_calls)

        span.add_event(event_name, attributes=attributes)
```

sdks/python/src/lilypad/_internal/otel/google/_patch.py
```diff
@@ -0,0 +1,199 @@
...
+    SyncStreamHandler,
+)
+from ..base.wrappers import (
+    AsyncStreamWrapper,
+    StreamWrapper,
+    create_async_wrapper,
+    create_sync_wrapper,
+)
+from ..utils import AsyncStreamProtocol, StreamProtocol
+from ._utils import (
```
willbakst: we should be namespacing `from . import _utils` to keep these private

sdks/python/src/lilypad/_internal/otel/google/_patch.py
```diff
@@ -0,0 +1,199 @@
...
+            attributes["gen_ai.content"] = content
+
+        tool_calls = message.get("tool_calls")
+        if tool_calls:
+            attributes["gen_ai.tool_calls"] = json.dumps(tool_calls)
+
+        span.add_event(event_name, attributes=attributes)
+
+
+def _process_response(
```
willbakst: nit: why not just implement `_utils.process_response`?

sdks/python/src/lilypad/_internal/otel/google/_patch.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def generate_content_stream_patch_factory(
+    tracer: Tracer,
+) -> SyncStreamHandler[P, GenerateContentResponse, GoogleMetadata, Models]:
+    def get_span_attributes_with_stream(
+        kwargs: dict[str, Any], client: Models | None, operation_name: str = "chat"
+    ) -> dict[str, AttributeValue]:
+        # Add stream=True to kwargs for streaming methods
+        kwargs_with_stream = {**kwargs, "stream": True}
```
willbakst: What is up with this? I don't understand this or the `force_stream` addition. Let's sync on this so I can understand what is going on here to properly comment.

sdks/python/src/lilypad/_internal/otel/google/_utils.py
```diff
@@ -0,0 +1,524 @@
...
+
+                buffer = buffers[index]
+
+                if isinstance(candidate.finish_reason, Enum):
+                    buffer.finish_reason = candidate.finish_reason.name
+
+                if candidate.content:
+                    content = candidate.content
+
+                    if content.parts:
```
willbakst: oof, this indentation is brutal. Let's sync on how we might be able to reduce the indentation here for improved readability.

sdks/python/src/lilypad/_internal/otel/google/_utils.py
```diff
@@ -0,0 +1,524 @@
...
+                                tool_buffer = buffer.tool_calls_buffers[tool_call_index]
+                                if tool_buffer and func_call.args:
+                                    arguments = json.dumps(func_call.args)
+                                    tool_buffer.append_arguments(arguments)
+
+
+def _extract_tools_from_config(
+    config: GenerateContentConfig, attributes: dict[str, AttributeValue]
+) -> None:
+    tools = config.tools
```
willbakst: nit: we can make this one line using the walrus operator

sdks/python/src/lilypad/_internal/otel/google/_utils.py
```diff
@@ -0,0 +1,524 @@
...
+    attributes: dict[str, AttributeValue] = {
+        gen_ai_attributes.GEN_AI_SYSTEM: "google_genai",
+        gen_ai_attributes.GEN_AI_REQUEST_MODEL: str(kwargs.get("model", "unknown")),
+        gen_ai_attributes.GEN_AI_OPERATION_NAME: operation_name,
+    }
+
+    if client:
+        set_server_address_and_port(client, attributes)
+
+    config = cast(
```
willbakst, line 83: same thing here re: unsafe case

sdks/python/src/lilypad/_internal/otel/google/_utils.py
```diff
@@ -0,0 +1,524 @@
...
+                )
+
+            if (frequency_penalty := config.frequency_penalty) is not None:
+                attributes[gen_ai_attributes.GEN_AI_REQUEST_FREQUENCY_PENALTY] = (
+                    frequency_penalty
+                )
+
+            _extract_tools_from_config(config, attributes)
+        elif isinstance(config, dict):
+            if (temperature := config.get("temperature")) and temperature is not None:
```
willbakst: checking `is not None` here seems redundant?

sdks/python/src/lilypad/_internal/otel/google/_utils.py
```diff
@@ -0,0 +1,524 @@
...
+                attributes[gen_ai_attributes.GEN_AI_REQUEST_PRESENCE_PENALTY] = (
+                    presence_penalty
+                )
+
+            if (frequency_penalty := config.frequency_penalty) is not None:
+                attributes[gen_ai_attributes.GEN_AI_REQUEST_FREQUENCY_PENALTY] = (
+                    frequency_penalty
+                )
+
+            _extract_tools_from_config(config, attributes)
```
willbakst: I'm not in love with this structure where we update `attributes` in the method rather than getting what we want and updating here.

Let's sync on how we can improve this.

sdks/python/src/lilypad/_internal/otel/google/_utils.py
```diff
@@ -0,0 +1,524 @@
...
+
+        if finish_reasons:
+            attributes[gen_ai_attributes.GEN_AI_RESPONSE_FINISH_REASONS] = tuple(
+                finish_reasons
+            )
+
+    span.set_attributes(attributes)
+
+
+def _process_content_to_messages(
```
willbakst: nit: we should name this `_convert_...` to match other methods that use the "convert" naming scheme

sdks/python/src/lilypad/_internal/otel/google/_utils.py
```diff
@@ -0,0 +1,524 @@
...
+    if text_parts:
+        message["content"] = "\n".join(text_parts)
+    if tool_calls:
+        message["tool_calls"] = tool_calls
+    if inline_data:
+        message["inline_data"] = inline_data
+    return message
+
+
+def process_response_to_events(
```
willbakst, line 262: I want to comb through this together and make sure we're matching the other providers as much as makes sense.

sdks/python/src/lilypad/_internal/otel/utils.py
```diff
@@ -29,6 +29,7 @@
 
 T = TypeVar("T")
 ChunkType = TypeVar("ChunkType")
```
willbakst: nit: these should be `ChunkT` and `ChunkCovariantT`?

also I feel like these are duplicates of existing type variables, may be worth unifying.

sdks/python/src/lilypad/_internal/otel/utils.py
```diff
@@ -41,43 +42,29 @@ class BaseMetadata(TypedDict, total=False):
...
         ...
 
-    @abstractmethod
-    async def aclose(self) -> None:
-        """Close the async stream and release resources."""
-        ...
-
 
 SyncAsyncStreamProtocol = TypeVar(
     "SyncAsyncStreamProtocol", bound=StreamProtocol | AsyncStreamProtocol
```
willbakst, line 83: <p class='graphite__hidden'><i>[Re: line -97]</i></p>
nit: we should unify these with the other types we recently created
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/670?utm_source=unchanged-line-comment">Graphite</a>.</p>

sdks/python/src/lilypad/_internal/otel/utils.py
```diff
@@ -41,43 +42,29 @@ class BaseMetadata(TypedDict, total=False):
     completion_tokens: int | None
 
 
-class StreamProtocol(Generic[ChunkType], ABC):
+class StreamProtocol(Protocol[ChunkType_co]):
```
willbakst: Hmm, are these true protocols? Makes sense given the name

sdks/python/src/lilypad/_internal/otel/utils.py
```diff
@@ -320,7 +308,8 @@ async def __aexit__(
 
     async def close(self) -> None:
         """Close the async stream and perform cleanup."""
-        await self.stream.aclose()
+        if aclose := getattr(self.stream, "aclose", None):
```
willbakst, line 288: when will it not have `aclose`?

sdks/python/src/lilypad/_internal/otel/google/_patch.py
willbakst: We should have a separate wrapper for stream methods instead of this.

Note that I believe Anthropic also has a `.stream` method that we should likely support, so we can use that + this as a means of understanding the best way to DRY our code.

willbakst: My feeling here is that we should just completely remove the stream wrapping here and then in a separate PR up stack handle stream alongside Anthropic so that we can solve both at the same time and properly refactor shared code.

sdks/python/src/lilypad/_internal/otel/google/_utils.py
willbakst: Here is an example that removes (I believe) 3 levels of indentation:

```python
def process_chunk(self, chunk: Any, buffers: list[ChoiceBuffer]) -> None:
    if not (candidate := chunk.candidate):
        return # pragma: no cover
    

    for candidate in chunk.candidates:
        index = candidate.index
        if index is None:
            index = 0

        while len(buffers) <= index:
            buffers.append(ChoiceBuffer(index=len(buffers)))

        buffer = buffers[index]

        if isinstance(candidate.finish_reason, Enum):
            buffer.finish_reason = candidate.finish_reason.name

        if not (content := candidate.content):
            continue  # pragma: no cover

        if not (parts := content.parts):
            continue  # pragma: no cover

        for part in content.parts:
            if part.text:
                buffer.append_text_content(part.text)

            if (function_call := part.function_call):
                tool_call_index = len(buffer.tool_calls_buffers)
                while len(buffer.tool_calls_buffers) <= tool_call_index:
                    buffer.tool_calls_buffers.append(None)
                if buffer.tool_calls_buffers[tool_call_index] is None:
                    tool_call_id = f"tool_{index}_{tool_call_index}"
                    function_name = (
                        function_call.name if function_call.name else ""
                    )
                    buffer.tool_calls_buffers[tool_call_index] = (
                        ToolCallBuffer(
                            index=tool_call_index,
                            tool_call_id=tool_call_id,
                            function_name=function_name,
                        )
                    )
                
                tool_buffer = buffer.tool_calls_buffers[tool_call_index]
                if tool_buffer and function_call.args:
                    arguments = json.dumps(function_call.args)
                    tool_buffer.append_arguments(arguments)
```

sdks/python/src/lilypad/_internal/otel/google/_utils.py
willbakst: My mistake, this is handling the `0` case for the `float` value

sdks/python/src/lilypad/_internal/otel/google/_utils.py
willbakst: For example, maybe this should be named `_extract_tool_names_from_config` and then return the extracted tool names

sdks/python/src/lilypad/_internal/otel/google/_utils.py
```diff
@@ -0,0 +1,524 @@
...
+                    presence_penalty
+                )
+
+            if (frequency_penalty := config.frequency_penalty) is not None:
+                attributes[gen_ai_attributes.GEN_AI_REQUEST_FREQUENCY_PENALTY] = (
+                    frequency_penalty
+                )
+
+            _extract_tools_from_config(config, attributes)
+        elif isinstance(config, dict):
```
willbakst: nit: this can technically just be an `else` clause given the cast. In these cases, we should either use `cast` because we are absolutely certain of the type (in which case we should use `if/else`) or we shouldn't cast and we should accept the `Any | None` type and then use `if/elif` where the final `else` clause is either dropped (i.e. do nothing) or handled with an error

sdks/python/src/lilypad/_internal/otel/google/_utils.py
```diff
@@ -0,0 +1,524 @@
...
+    }
+
+    if client:
+        set_server_address_and_port(client, attributes)
+
+    config = cast(
+        GenerateContentConfig | GenerateContentConfigDict, kwargs.get("config")
+    )
+    if config:
+        if isinstance(config, GenerateContentConfig):
```
willbakst, line 87: nit: I wonder if we could do something like dump this so that it acts more like the dictionary counterpart so we can reduce the duplication here to a single code path

sdks/python/src/lilypad/_internal/otel/google/_utils.py
willbakst: After taking a deeper look, I think we should unify all providers to have a single `_utils.process_response` method that can then use whatever private methods internally that it needs to do the processing.

This way all providers will have a similar utility interface unified on `process_response`. It's nice to ensure the `_utils` private package operates similarly across the various providers as it will be easier to reason about / understand / maintain.

It's not that the logic is shared, it's that the public interface is matching.

willbakst: To be more clear, we should standardize each provider's `_utils.py` module to export the same named methods, and those named methods should implement the protocols.

For example, `GetSpanAttributes` would have a corresponding `_utils.get_span_attributes` method implemented for every provider. We can then use whatever private utility methods we want to implement this shared interface, but each provider will have a `_utils` module with a matching interface that makes it easier to reason about.

sdks/python/src/lilypad/_internal/otel/utils.py
willbakst: Note: this is for Google

sdks/python/src/lilypad/_internal/otel/google/_patch.py
```diff
@@ -0,0 +1,111 @@
...
+    ContentListUnionDict,
+    ContentUnionDict,
+)
+from google.genai.models import AsyncModels, Models
+from google.genai.types import GenerateContentResponse
+
+P = ParamSpec("P")
+
+
+def _process_messages(
```
willbakst, line 36: I have an idea for a refactor where we can unify all of the provider-specific utility interfaces to implement an abstract base class for consistency. We can sync on this online.

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -36,3 +46,29 @@ class LLMOpenTelemetryMessage(TypedDict, total=False):
 
     tool_calls: list[LLMOpenTelemetryToolCall]
     """The list of tool calls, if any."""
+
+    inline_data: list[LLMOpenTelemetryInlineData]
+    """The list of data to be embedded in the message, if any."""
+
+
+class ToolCallFunctionProtocol(Protocol):
```
willbakst, line 54: These seem like duplicates of the existing protocols above?

sdks/python/src/lilypad/_internal/otel/google/_utils.py
```diff
@@ -0,0 +1,313 @@
...
+    attributes: dict[str, AttributeValue] = {}
+
+    if response.model_version:
+        attributes[gen_ai_attributes.GEN_AI_RESPONSE_MODEL] = response.model_version
+
+    if response.response_id:
+        attributes[gen_ai_attributes.GEN_AI_RESPONSE_ID] = response.response_id
+
+    if response.usage_metadata:
+        usage = response.usage_metadata
```
willbakst, line 130: This and other places can be walrus

sdks/python/src/lilypad/_internal/otel/google/_utils.py
```diff
@@ -0,0 +1,313 @@
...
+    output_tokens: int
+
+
+def _extract_tool_names_from_config(
+    config: GenerateContentConfigDict,
+) -> list[str] | None:
+    if not (tools := config.get("tools")):
+        return None
+
+    function_names: list[str] = []
```
willbakst, line 57: nit: we should name this `tool_names`

sdks/python/src/lilypad/_internal/otel/google/_utils.py
```diff
@@ -0,0 +1,313 @@
...
+            continue  # pragma: no cover
+
+        message: LLMOpenTelemetryMessage = {
+            "role": content.role or "model",
+        }
+
+        text_parts: list[str] = []
+        tool_calls: list[LLMOpenTelemetryToolCall] = []
+
+        if content.parts:
```
willbakst, line 277: nit: this should use continue to remove one more layer of indentation

willbakst (commented):
We should go through and finalize this with some pair programming to make sure we're on the same page since there are some comments here that haven't been addressed properly.

PR #678, opened 2025-08-20
Implement Python OpenTelemetry Exporters

koxudaxi:


koxudaxi:
* **#694** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/694?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#695](https://github.com/Mirascope/lilypad/pull/695) <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/695?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#696](https://github.com/Mirascope/lilypad/pull/696) <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/696?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#693** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/693?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#692** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/692?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#691** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/691?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#679** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/679?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#678** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/678?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/678?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#689** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/689?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#690](https://github.com/Mirascope/lilypad/pull/690) <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/690?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `v1`


<h2></h2>

sdks/python/src/lilypad/__init__.py
```diff
@@ -7,6 +7,7 @@
 from .client import (
     Lilypad,
     AsyncLilypad,
+    get_client,
```
willbakst: note: in line with other comments, we should be mindful of what we publicly export

sdks/python/src/lilypad/_internal/otel/exporters/config.py
willbakst, line 1: nit: should we rename this to `configuration.py` to better match our other `configuration.py` module naming?

sdks/python/src/lilypad/_internal/otel/exporters/config.py
```diff
@@ -0,0 +1,39 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from lilypad.client import Lilypad
+from .processors import LLMSpanProcessor
+
+
+@dataclass
+class ConfigureExportersConfig:
```
willbakst: What is the benefit of having this dataclass vs. just putting each field as an argument of the `configure_exporters` method? Seems like an unnecessary indirection?

sdks/python/src/lilypad/_internal/otel/exporters/exporters.py
```diff
@@ -0,0 +1,123 @@
...
+        This method is called by BatchSpanProcessor with a batch
+        of completed spans to export.
+
+        Args:
+            spans: Sequence of completed spans to export.
+
+        Returns:
+            SpanExportResult indicating success or failure.
+        """
+        # Delegate to transport which handles OTLP conversion
```
willbakst: nit: this comment (and other similar ones) seem unnecessary? If necessary, why not just include it in the docstring?

sdks/python/src/lilypad/_internal/otel/exporters/otlp_protocols.py
willbakst, line 1: nit: is this only for `OTLP` type-safety? Why not just use the open-telemetry types directly? I feel like we use those for the LLM instrumentation, so I'm not sure why we're duplicating?

sdks/python/src/lilypad/_internal/otel/exporters/processors.py
```diff
@@ -0,0 +1,121 @@
...
+
+from opentelemetry.context import Context
+from opentelemetry.sdk.trace import SpanProcessor, ReadableSpan
+from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+from .types import SpanStartEvent
+from .exporters import ImmediateStartExporter
+
+
+class LLMSpanProcessor(SpanProcessor):
```
willbakst: Why is this specifically for LLM span processing? Shouldn't this just be `LilypadSpanProcessor`? Won't this handle spans that are just normal spans too that have nothing to do with LLMs?

sdks/python/src/lilypad/_internal/otel/exporters/processors.py
```diff
@@ -0,0 +1,121 @@
...
+        via the start exporter in a non-blocking manner.
+
+        Args:
+            span: The span that just started.
+            parent_context: Optional parent context for the span.
+        """
+        if self._shutdown:
+            return
+
+        # Note: _create_start_event is not implemented yet
```
willbakst: What will `_create_start_event` do differently from this?

sdks/python/src/lilypad/_internal/otel/exporters/transport.py
```diff
@@ -0,0 +1,273 @@
...
+events to the Lilypad ingestion endpoint. It wraps the Fern-generated
+LilypadClient to provide the interface needed by OpenTelemetry exporters.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+from dataclasses import dataclass
+
+from lilypad._generated import (
```
willbakst: we should be using relative imports for our own code? Please do a sweep to fix any and all instances of this

sdks/python/src/lilypad/_internal/otel/exporters/transport.py
```diff
@@ -0,0 +1,273 @@
...
+    ScopeSpans,
+)
+from .otlp_protocols import AttributeValueType
+
+from lilypad._generated.client import Lilypad  # noqa: PLC2701
+from opentelemetry.sdk.trace import ReadableSpan
+from opentelemetry.sdk.trace.export import SpanExportResult
+
+
+class TransportError(Exception):
```
willbakst: Should we have a `LilypadException` we use as the base for more easily handling Lilypad-specific exceptions? For example, when we want to ensure that we can call a user function safely even if our code errors, but we want to make sure we properly bubble up user errors to the user?

sdks/python/src/lilypad/_internal/otel/exporters/transport.py
```diff
@@ -0,0 +1,273 @@
...
+    """Transport layer for OpenTelemetry spans using Fern client.
+
+    This transport accepts standard OpenTelemetry ReadableSpan objects
+    and converts them to OTLP format for transmission via the Fern client.
+    """
+
+    def __init__(
+        self,
+        client: Lilypad,
+        config: TelemetryConfig | None = None,
```
willbakst: Again, what's the benefit of this config approach? Is this something that comes from OTel?

sdks/python/src/lilypad/_internal/otel/exporters/transport.py
```diff
@@ -0,0 +1,273 @@
...
+            ),
+            name=span.name,
+            kind=span.kind.value if span.kind else 0,
+            start_time_unix_nano=str(span.start_time) if span.start_time else "0",
+            end_time_unix_nano=str(span.end_time) if span.end_time else "0",
+            attributes=attributes,
+            status=status,
+        )
+
+    def _convert_attribute_value(self, value: AttributeValueType) -> KeyValueValue:
```
willbakst: Why not use `from opentelemetry.util.types import AttributeValue` ??

We should use their types directly, no? Why reimplement them?

sdks/python/src/lilypad/_internal/otel/exporters/types.py
```diff
@@ -0,0 +1,121 @@
...
+immediate start event transmission and batched end event export.
+"""
+
+from enum import Enum
+from typing import TypedDict, Literal
+
+from opentelemetry.util.types import AttributeValue
+
+
+SpanKindLiteral = Literal["CLIENT", "SERVER", "PRODUCER", "CONSUMER", "INTERNAL"]
```
willbakst: I feel like this should just be `SpanKind`? Why include `Literal` in the name?

Below you have `StatusCode`, which feels more correct than `StatusCodeLiteral`, so let's do the same here

sdks/python/src/lilypad/_internal/otel/exporters/types.py
```diff
@@ -0,0 +1,121 @@
...
+StatusCode = Literal["UNSET", "OK", "ERROR"]
+
+
+class SpanEventType(Enum):
```
willbakst: should we make this `class SpanEventType(str, Enum): ...`?

sdks/python/src/lilypad/_internal/otel/exporters/types.py
```diff
@@ -0,0 +1,121 @@
...
+    SPAN_UPDATED = "span_updated"
+    SPAN_COMPLETED = "span_completed"
+
+
+class SpanEvent(TypedDict):
+    """Individual event within a span."""
+
+    name: str
+    timestamp: int
+    """Nanoseconds since epoch (OTel standard)."""
```
willbakst, line 33: let's put docstrings on all of the fields with newline separation like I do for the `otel/types.py` module. Also, I feel like there's a fair bit of overlap with many of the types I defined? We should probably unify them? We can do that in a separate unification PR rather than here, but I wanted to note it while I'm seeing it.

sdks/python/src/lilypad/_internal/otel/exporters/types.py
```diff
@@ -0,0 +1,121 @@
...
+    This event is sent immediately when a span starts to provide
+    real-time visibility into long-running operations.
+    """
+
+    trace_id: str
+    span_id: str
+    parent_span_id: str | None
+    name: str
+    start_time: int
+    """Nanoseconds since epoch."""
```
willbakst, line 95: again, weird where only some have docstrings, please put docstrings on all of them with newline separation.

sdks/python/src/lilypad/_internal/otel/exporters/types.py
```diff
@@ -0,0 +1,121 @@
...
+    links: list[Link]
+
+
+class GenAIEventNames:
+    """Standard event names for GenAI spans.
+
+    These follow OpenTelemetry semantic conventions for GenAI.
+    """
+
+    PROMPT = "gen_ai.content.prompt"
```
willbakst: Aren't there specific constants in the otel library we can use like we do in the `otel/` module rather than using these direct strings? Feels better to use their constants?

sdks/python/src/lilypad/configuration.py
```diff
@@ -38,16 +45,25 @@ def configure(
     log_handlers: list[logging.Handler] | None = None,
     log_format: str | None = None,
     log_level: int = DEFAULT_LOG_LEVEL,
+    lilypad_endpoint: str | None = None,
```
willbakst: Why make this something a user can configure? Won't it be fixed to our specific endpoint?

sdks/python/src/lilypad/configuration.py
```diff
@@ -38,16 +45,25 @@ def configure(
...
 
     Args:
-        log_handlers (list[logging.Handler]): Custom logging handlers to use. If None, defaults to RichHandler.
-        log_format (str): Custom logging format. If None, defaults to a standard format.
-        log_level (int): The logging level to set. Defaults to logging.INFO.
+        log_handlers: Custom logging handlers to use. If None, defaults to RichHandler.
+        log_format: Custom logging format. If None, defaults to a standard format.
+        log_level: The logging level to set. Defaults to logging.INFO.
+        lilypad_endpoint: Optional Lilypad ingestion endpoint URL for tracing.
+        api_key: Optional API key for authentication with Lilypad endpoint.
```
willbakst: What is `lilypad.configure` retrieved settings from the settings context manager? Then a user could set the API key and Base URL with something like `with lilypad.settings(...): lilypad.configure(...)`

sdks/python/src/lilypad/configuration.py
```diff
@@ -59,3 +75,32 @@ def configure(
     if trace.get_tracer_provider().__class__.__name__ == "TracerProvider":
         logger.info("TracerProvider already initialized.")  # noqa: T201
         return
+
+    if lilypad_endpoint and api_key:
```
willbakst: Seems weird to me that `lilypad_endpoint` is optional but then nothing happens if not set?

sdks/python/src/lilypad/_internal/otel/exporters/otlp_protocols.py
koxudaxi: make sense. I will delete it.

sdks/python/src/lilypad/_internal/otel/exporters/transport.py
koxudaxi: I'm using OpenTelemetry's AttributeValue type directly. I import it and use it as the input type. The _convert_attribute_value method isn't reimplementing OpenTelemetry types. It's converting from OpenTelemetry's AttributeValue format to the Fern-generated KeyValueValue format that the Lilypad API expects.

sdks/python/src/lilypad/_internal/otel/exporters/types.py
koxudaxi: You're right to question this. I checked the OpenTelemetry semantic conventions - they define attribute constants like GEN_AI_PROMPT and GEN_AI_COMPLETION but don't  specify standard event names for GenAI spans.
These event names (gen_ai.content.prompt, etc.) are custom for Lilypad's telemetry. They're used when calling span.add_event() to mark specific moments in the span  lifecycle.
I've updated the docstring to clarify these are custom event names, not OpenTelemetry standards. The conventions focus on attributes and operation names, but leave event naming up to implementations.

sdks/python/src/lilypad/configuration.py
koxudaxi: It was intended to support self-hosted deployment.

sdks/python/src/lilypad/configuration.py
koxudaxi: Good idea! Updated configure() to retrieve settings from the context manager. 
Now users can do:
```
with lilypad.settings(api_key="your-api-key"):
    lilypad.configure()  # Uses API key from context
```

Also:
- Removed api_key parameter from configure()
- Renamed lilypad_endpoint to telemetry_endpoint for clarity
- Exported settings from main package
The telemetry endpoint is still configurable for testing/on-prem deployments, but most users won't need to set it.

sdks/python/src/lilypad/configuration.py
koxudaxi: I removed it.

sdks/python/src/lilypad/__init__.py
```diff
@@ -13,6 +14,7 @@
     from ._internal.otel import instrument_anthropic
 
 __all__ = [
+    "LilypadException",
```
willbakst, line 20: We also need to export `Lilypad` and `AsyncLilypad`, right? So that users can use those clients we've implemented?

sdks/python/src/lilypad/__init__.py
```diff
@@ -4,6 +4,7 @@
 
 from . import configuration
 from .configuration import configure
+from ._internal.errors import LilypadException
```
willbakst: Why export only `LilypadException`? Do we expect users only to ever need that one?

sdks/python/src/lilypad/configuration.py
```diff
@@ -4,10 +4,15 @@
 from secrets import token_bytes
 
 from opentelemetry import trace
+from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.sdk.trace.id_generator import IdGenerator
 from opentelemetry.trace import INVALID_SPAN_ID, INVALID_TRACE_ID
 from rich.logging import RichHandler
 
+from ._internal.otel.exporters import configure_exporters
```
willbakst: So `configure_exporters` and `get_settings` are now public, when they likely shouldn't be?

sdks/python/src/lilypad/configuration.py
willbakst: Right, but `base_url` should already cover that, right?

sdks/python/src/lilypad/_internal/otel/exporters/configuration.py
```diff
@@ -0,0 +1,40 @@
...
+
+from ....client import Lilypad
+from .constants import (
+    DEFAULT_MAX_RETRY_ATTEMPTS,
+    DEFAULT_TIMEOUT,
+)
+from .processors import LilypadSpanProcessor
+
+
+def configure_exporters(
```
willbakst: Do we ever use this anywhere except for the main `lilypad.configure` method? I wonder if this should be a private method in that file rather than live here?

sdks/python/src/lilypad/_internal/otel/exporters/constants.py
```diff
@@ -0,0 +1,7 @@
+"""Constants for the Lilypad exporters."""
+
+from __future__ import annotations
+
+DEFAULT_EXPORT_BASE_URL = "https://lilypad-api.mirascope.com/v1"
```
willbakst: This is the wrong URL? Should be `https://v1.lilypad.mirascope.com/api/v1`?

sdks/python/src/lilypad/_internal/otel/exporters/transport.py
```diff
@@ -0,0 +1,254 @@
...
+
+        status = None
+        if span.status:
+            status = SpanStatus(
+                code=span.status.status_code.value,
+                message=span.status.description,
+            )
+
+        context = span.get_span_context()
+        if not context:
```
willbakst: Shouldn't we fail if there's no context? What happens if there's no context? Can we actually handle that case?

sdks/python/src/lilypad/_internal/otel/exporters/types.py
willbakst: So we don't forget, can you please add a TODO here to do the unification?

Something like `# TODO: unify/DRY types in the _internal package`

sdks/python/src/lilypad/_internal/otel/exporters/types.py
willbakst: I don't see where you're using these, so I'm having a hard time following. Can you point out where this is being used so I can better understand why we're doing this?

koxudaxi: This was a definition for code that had already been deleted, so I deleted the definition along with it.

sdks/python/src/lilypad/configuration.py
koxudaxi: yes, I deleted it.

willbakst (approved):

willbakst:
### Merge activity

* **Sep 10, 4:05 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/lilypad/678).

PR #680, opened 2025-08-21
refactor: OpenAI and Anthropic OpenTelemetry instrumentation

willbakst:
Large refactor of the OpenAI and Anthropic OpenTelemetry instrumentation code.

The primary goal was maintainability through simplicity.

For now, I've added a few `# pragma: no cover` lines to `base_instrumentor.py` and `stream_wrapper.py` that we can revisit. It's mostly stuff like exception handling that isn't fully covered yet.


willbakst:
* **#686** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/686?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#685** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/685?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#684** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/684?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#681** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/681?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#680** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/680?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/680?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v1`


<h2></h2>

sdks/python/src/lilypad/_internal/otel/openai.py
```diff
@@ -0,0 +1,303 @@
...
+        service_tier = kwargs.get("service_tier")
+        if service_tier == "auto":
+            service_tier = None
+        return GenAIRequestAttributes(
+            GEN_AI_SYSTEM=OPENAI_SYSTEM,
+            SERVER_ADDRESS=client._client.base_url.host,
+            SERVER_PORT=client._client.base_url.port,
+            GEN_AI_REQUEST_MODEL=kwargs.get("model"),
+            GEN_AI_REQUEST_TEMPERATURE=kwargs.get("temperature"),
+            GEN_AI_REQUEST_TOP_P=kwargs.get("p") or kwargs.get("top_p"),
```
koxudaxi: `kwargs ` doesn't have `"p"`

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -34,5 +148,324 @@ class LLMOpenTelemetryMessage(TypedDict, total=False):
...
+            "timestamp": self.timestamp,
+        }
+
+
+@dataclass(kw_only=True)
+class SystemMessageEvent(BaseMessageEvent):
+    """[MISSING DOCSTRING]"""
+
+    @property
+    def role(self) -> str:
```
koxudaxi: Should be `Literal["system"]`  as return type?

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -34,5 +148,324 @@ class LLMOpenTelemetryMessage(TypedDict, total=False):
...
+    content: Any | None = None
+    """[MISSING DOCSTRING]"""
+
+
+@dataclass(kw_only=True)
+class UserMessageEvent(BaseMessageEvent):
+    """[MISSING DOCSTRING]"""
+
+    @property
+    def role(self) -> str:
```
koxudaxi: Should be `Literal["user"]`  as return type?

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -34,5 +148,324 @@ class LLMOpenTelemetryMessage(TypedDict, total=False):
...
+    content: Any | None = None
+    """[MISSING DOCSTRING]"""
+
+
+@dataclass(kw_only=True)
+class AssistantMessageEvent(BaseMessageEvent):
+    """[MISSING DOCSTRING]"""
+
+    @property
+    def role(self) -> str:
```
koxudaxi: Should be `Literal["assistant"]`  as return type?

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -34,5 +148,324 @@ class LLMOpenTelemetryMessage(TypedDict, total=False):
...
+    tool_calls: list[ToolCall] | None = None
+    """[MISSING DOCSTRING]"""
+
+
+@dataclass(kw_only=True)
+class ToolMessageEvent(BaseMessageEvent):
+    """[MISSING DOCSTRING]"""
+
+    @property
+    def role(self) -> str:
```
koxudaxi: Should be `Literal["tool"]`  as return type?

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -34,5 +148,324 @@ class LLMOpenTelemetryMessage(TypedDict, total=False):
...
+        for field in fields(self):
+            if field.name == "system":
+                continue
+            value = getattr(self, field.name)
+            if not value and value != 0:
+                continue
+            attributes[field.name] = (
+                value
+                if isinstance(value, str | int | float | bool)
+                else json_dumps(value)
```
koxudaxi: Should we define `PrimitiveType` and a function for serialization?

sdks/python/src/lilypad/_internal/otel/base_instrumentor.py
```diff
@@ -0,0 +1,234 @@
...
+    @overload
+    def _wrap(
+        self,
+        method: BoundAsyncMethod[P, ResponseT],
+        wrapper: AsyncMethodWrapper[P, ResponseT, ClientT],
+    ) -> None:
+        """[MISSING DOCSTRING]"""
+        ...
+
+    def _wrap(
```
koxudaxi, line 86: `_wrap`  doesn't use `self`. Do we need `@classmethod` or `@staticmethod`?

sdks/python/src/lilypad/_internal/otel/base_instrumentor.py
```diff
@@ -0,0 +1,234 @@
...
+            provider_kwargs = cast(KwargsT, kwargs)
+            request_attributes = self._get_request_attributes(provider_kwargs, client)
+            with self._span(request_attributes) as span:
+                if span.is_recording():
+                    for event in self._process_messages(provider_kwargs):
+                        span.add_event(**event.dump())
+                try:
+                    response = wrapped(*args, **kwargs)
+                    if handle_stream and (
+                        kwargs.get("stream") or isinstance(response, Stream)
```
koxudaxi: If the return value type is `Stream`, is it unnecessary to check `"stream"` in `kwargs`?

sdks/python/src/lilypad/_internal/otel/base_instrumentor.py
```diff
@@ -0,0 +1,234 @@
...
+            provider_kwargs = cast(KwargsT, kwargs)
+            request_attributes = self._get_request_attributes(provider_kwargs, client)
+            with self._span(request_attributes) as span:
+                if span.is_recording():
+                    for event in self._process_messages(provider_kwargs):
+                        span.add_event(**event.dump())
+                try:
+                    response = await wrapped(*args, **kwargs)
+                    if handle_stream and (
+                        kwargs.get("stream") or isinstance(response, AsyncStream)
```
koxudaxi: same above

sdks/python/src/lilypad/_internal/otel/base_instrumentor.py
```diff
@@ -0,0 +1,234 @@
...
+                    ):
+                        async_stream_wrapper = AsyncStreamWrapper[StreamChunkT](
+                            span,
+                            cast(AsyncStream, response),
```
koxudaxi: same above

sdks/python/src/lilypad/_internal/otel/base_instrumentor.py
```diff
@@ -0,0 +1,234 @@
...
+                    for event in self._process_messages(provider_kwargs):
+                        span.add_event(**event.dump())
+                try:
+                    response = wrapped(*args, **kwargs)
+                    if handle_stream and (
+                        kwargs.get("stream") or isinstance(response, Stream)
+                    ):
+                        stream_wrapper = StreamWrapper[StreamChunkT](
+                            span,
+                            cast(Stream, response),
```
koxudaxi: If we remove the `kwargs.get("stream")` by the previous comment, can we remove the cast.

sdks/python/src/lilypad/_internal/otel/base_instrumentor.py
```diff
@@ -0,0 +1,234 @@
...
+    ) -> tuple[list[ChoiceEvent], GenAIResponseAttributes]:
+        """[MISSING DOCSTRING]"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def _process_chunk(
+        self,
+        chunk: StreamChunkT,
+    ) -> tuple[GenAIResponseAttributes, list[ChoiceDelta]]:
+        """[MISSING DOCSTRINg]"""
```
koxudaxi: DOCSTRING

sdks/python/src/lilypad/_internal/otel/anthropic.py
```diff
@@ -0,0 +1,318 @@
...
+) -> None:
+    """[MISSING DOCSTRING]"""
+    if _utils.client_is_already_instrumented(client):
+        return
+
+    instrumentor = AnthropicInstrumentor()
+
+    if isinstance(client, Anthropic | AnthropicBedrock | AnthropicVertex):
+        instrumentor.instrument_generate(client.messages.create, handle_stream=True)
+        instrumentor.instrument_generate(client.messages.stream, handle_stream=True)
```
koxudaxi: I think the `.stream` method return `MessageStreamManager`.
`_process_assistant_message` can't process the object.

```
Traceback (most recent call last):
  File "/Users/koudai/PycharmProjects/wd/lilypad/sdks/python/examples/instrument_anthropic_with_console.py", line 20, in <module>
    with client.messages.stream(
         ~~~~~~~~~~~~~~~~~~~~~~^
        max_tokens=1024,
        ^^^^^^^^^^^^^^^^
        messages=[{"role": "user", "content": "Hello"}],
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        model="claude-opus-4-1-20250805",
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ) as stream:
    ^
  File "/Users/koudai/PycharmProjects/wd/lilypad/sdks/python/src/lilypad/_internal/otel/base_instrumentor.py", line 133, in wrapper
    choice_events, response_attributes = self._process_response(
                                         ~~~~~~~~~~~~~~~~~~~~~~^
        response
        ^^^^^^^^
    )
    ^
  File "/Users/koudai/PycharmProjects/wd/lilypad/sdks/python/src/lilypad/_internal/otel/anthropic.py", line 218, in _process_response
    content, tool_calls = _process_assistant_message(response)
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/Users/koudai/PycharmProjects/wd/lilypad/sdks/python/src/lilypad/_internal/otel/anthropic.py", line 119, in _process_assistant_message
    content = message["content"]
              ~~~~~~~^^^^^^^^^^^
TypeError: 'MessageStreamManager' object is not subscriptable


```

sdks/python/src/lilypad/_internal/otel/anthropic.py
```diff
@@ -0,0 +1,318 @@
...
+    else:
+        instrumentor.instrument_async_generate(
+            client.messages.create, handle_stream=True
+        )
+        instrumentor.instrument_async_generate(
+            client.messages.stream, handle_stream=True
```
koxudaxi: same above

sdks/python/src/lilypad/_internal/otel/openai.py
```diff
@@ -0,0 +1,303 @@
...
+        return
+
+    instrumentor = _OpenAIInstrumentor()
+
+    if isinstance(client, OpenAI | AzureOpenAI):
+        instrumentor.instrument_generate(
+            client.chat.completions.create, handle_stream=True
+        )
+        instrumentor.instrument_generate(
+            client.chat.completions.stream, handle_stream=True
```
koxudaxi: I think the `.stream` method return `ChatCompletionStreamManager`.
`_process_response`can't process the object.
```
Traceback (most recent call last):
  File "/Users/koudai/PycharmProjects/wd/lilypad/sdks/python/examples/instrument_openai_with_console.py", line 19, in <module>
    with client.chat.completions.stream(
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        model="gpt-4o-mini",
        ^^^^^^^^^^^^^^^^^^^^
        messages=[{"role": "user", "content": "Hello!"}],
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ) as stream:
    ^
  File "/Users/koudai/PycharmProjects/wd/lilypad/sdks/python/src/lilypad/_internal/otel/base_instrumentor.py", line 133, in wrapper
    choice_events, response_attributes = self._process_response(
                                         ~~~~~~~~~~~~~~~~~~~~~~^
        response
        ^^^^^^^^
    )
    ^
  File "/Users/koudai/PycharmProjects/wd/lilypad/sdks/python/src/lilypad/_internal/otel/openai.py", line 211, in _process_response
    for choice in response.choices:
                  ^^^^^^^^^^^^^^^^
AttributeError: 'ChatCompletionStreamManager' object has no attribute 'choices'

```

sdks/python/tests/lilypad/_internal/otel/test_anthropic.py
```diff
@@ -0,0 +1,515 @@
+"""Tests for Anthropic instrumentation."""
```
koxudaxi, line 1: We need `.stream()` tests

sdks/python/tests/lilypad/_internal/otel/test_openai.py
```diff
@@ -0,0 +1,459 @@
+"""Tests for OpenAI instrumentation."""
```
koxudaxi, line 1: We need `.stream()` tests

koxudaxi (commented):
Looks good.
I left few comments.

sdks/python/src/lilypad/_internal/otel/openai.py
willbakst: Good catch!

sdks/python/src/lilypad/_internal/otel/types.py
willbakst: Good catch! Done.

sdks/python/src/lilypad/_internal/otel/types.py
willbakst: Would you consider `list` or `set` or `dict` a primitive type or just these four?

The reason for this conditional is that if I just call `json_dumps(value)` then something like a value of `0` becomes `'0'` instead of just the raw integer, which is incorrect. I wonder if we could update `json_dumps` to handle this internally?

WDYT?

@teamdandelion since Koudai is currently offline, if you can chime in to help resolve this that would be great!

sdks/python/src/lilypad/_internal/otel/base_instrumentor.py
willbakst: Yeah, I think `@staticmethod` makes the most sense. Has slight performance benefits and makes it clear that the method should not depend on the state of the class.

I've made the same update to the `@abstractmethod` methods so that it's clear when implementing the class that those methods must not rely on the state of the class.

sdks/python/src/lilypad/_internal/otel/base_instrumentor.py
willbakst: Good catch. Plan is for Claude to sweep through and write docstrings once finalized in a following stack item.

sdks/python/tests/lilypad/_internal/otel/test_openai.py
willbakst: We don't yet support `.stream()` in this PR because that involves the context manager. I'm planning on addressing this in a following stack item. I've commented out the instrumentation in the meantime to make that more clear.

sdks/python/src/lilypad/_internal/otel/anthropic.py
willbakst: Yeah, this PR doesn't handle that yet. As mentioned in my other comment, I've commented out this code with TODOs to note that we need to add handling that doesn't exist yet.

sdks/python/tests/lilypad/_internal/otel/test_anthropic.py
willbakst: See other comment.

sdks/python/src/lilypad/_internal/otel/base_instrumentor.py
willbakst: After some checks, it appearas the `isinstance(response, Stream)` check doesn't really work, but `kwargs.get("stream")` works and we no longer need `handle_stream`, so I've removed it. Same below.

sdks/python/src/lilypad/_internal/otel/base_instrumentor.py
willbakst: As I mentioned in my other comment, I couldn't get `isinstance(response, Stream)` to work, so I've removed that, and we still need to cast.

I am struggling to think of a better way off the top of my head, so I'm going to leave a TODO to revisit.

sdks/python/src/lilypad/_internal/otel/openai.py
willbakst: Yeah as mentioned in my other comment, I've commented this out and added TODOs to handle this in following work.

sdks/python/src/lilypad/_internal/otel/base_instrumentor.py
willbakst: I was actually able to get this working. Good call! I've removed the cast for `Stream` and `AsyncStream` and removed the `kwargs.get("stream")` check as well so we just check the type of the response.

sdks/python/src/lilypad/_internal/otel/base_instrumentor.py
willbakst: I actually got this working. Good call!

.codespellrc
```diff
@@ -1,3 +1,3 @@
 [codespell]
 skip = node_modules,bun.lock,dist,coverage,**/cassettes/**,htmlcov
-ignore-words-list = afterAll,assistin
+ignore-words-list = afterAll,assistin,speakin
```
willbakst: @teamdandelion do you think we should ignore the words that come up from the VCR tests, update the tests not to use pirate speak, or ignore the test files themselves?

I'm currently leaning towards updating not to use pirate speak lol.

teamdandelion: lol maybe just use a codespell inline ignore?
`fer when yer AI be speakin like a pirate?  # codespell:ignore fer yer speakin`

sdks/python/src/lilypad/_internal/otel/stream_wrappers.py
```diff
@@ -0,0 +1,358 @@
...
+        process_chunk: ProcessChunk[ChunkT],
+    ) -> None:
+        """Initialize the stream wrapper with telemetry components."""
+        self.span = span
+        self.stream = stream
+        self.process_chunk = process_chunk
+        self.response_attributes = GenAIResponseAttributes()
+        self.choice_buffers: list[ChoiceBuffer] = []
+        self._span_started = False
+        self._setup()
```
teamdandelion: Looks like this is is duplicated between StreamWrapper and AsyncStreamWrapper, except for the type hint on stream. Wondering if we can move the implementation to BaseStreamWrapper with an extra type variable for Stream vs AsyncStream? Or does that run into issues with the fact that Stream/AsyncStream are themselves generic on ChunkT?

sdks/python/src/lilypad/_internal/otel/stream_wrappers.py
```diff
@@ -0,0 +1,358 @@
...
+            return chunk
+        except StopAsyncIteration:
+            self._cleanup()
+            raise
+        except Exception as error:  # pragma: no cover
+            self.span.set_status(Status(StatusCode.ERROR, str(error)))
+            self.span.set_attribute(
+                error_attributes.ERROR_TYPE, type(error).__qualname__
+            )
+            self._cleanup()
```
teamdandelion, line 357: I'm noting this logic is duplicated 2x per class, wonder if this could also become a helper method on the base class

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -34,5 +148,324 @@ class LLMOpenTelemetryMessage(TypedDict, total=False):
...
+        """[MISSING DOCSTRING]"""
+        attributes: dict[str, AttributeValue] = {
+            gen_ai_attributes.GEN_AI_SYSTEM: self.system
+        }
+        for field in fields(self):
+            if field.name == "system":
+                continue
+            value = getattr(self, field.name)
+            if not value and value != 0:
+                continue
```
teamdandelion: We're skipping empty string and `False`? Why is that?

sdks/python/src/lilypad/_internal/otel/types.py
teamdandelion: Hmm, here's the AttributeValue type from otel:

```py
AttributeValue = Union[
    str,
    bool,
    int,
    float,
    Sequence[str],
    Sequence[bool],
    Sequence[int],
    Sequence[float],
]
```

As written the code would convert `[True]` into `'[true]'` which isn't right - `[True]` was directly representable as an AttributeValue without string encoding. 

Here's an (untested) function which I believe would always use the AttributeValue representation if possible, and then fall back to using json.dumps only if the data is not representable as such. Wdyt?

```py
def encode_attr_value(input: AnyValue) -> AttributeValue:
    if isinstance(input, (str, bool, int, float)):
        return input 
    if isinstance(input, list):
        if not input:
            return input 
        item_types = list({type(item) for item in input})
        if len(item_types) == 1 and item_types[0] in (str, bool, int, float):
            return input 
    # Not directly representable as an AttributeValue, so use the JSON encoding
    return json.dumps(input) 
```

sdks/python/tests/lilypad/_internal/otel/test_anthropic.py
```diff
@@ -0,0 +1,515 @@
...
+                    "attributes": {
+                        "gen_ai.system": "anthropic",
+                        "content": "Hello, say 'Hi' back to me",
+                    },
+                },
+                {
+                    "name": "gen_ai.choice",
+                    "attributes": {
+                        "gen_ai.system": "anthropic",
+                        "message": '{"role":"assistant","content":"Ahoy, matey! Ye be speakin\' to the most scurvy-ridden pirate of the high seas. Avast, ye landlubber! *clears pirate-y throat* Ah"}',
```
teamdandelion: Try adding:  `# codespell:ignore speakin`

.codespellrc
willbakst: Good call. This worked.

sdks/python/src/lilypad/_internal/otel/types.py
willbakst: Hmm, technically there are no fields of these types, so shouldn't be an issue, but I can be more explicit with instance checking.

I'm meaning to do a fuller sweep through all of these types in a following PR to see where I can further unify these dump methods for `SpanEvent`. Things like organization of variables, naming, etc. I want to sweep through for sure.

sdks/python/src/lilypad/_internal/otel/types.py
willbakst: Ah yeah this is a great catch. Definitely needs to be updated. Let me take another look with your suggestion and comment when I've resolved it.

sdks/python/src/lilypad/_internal/otel/stream_wrappers.py
willbakst: Ok I've refactored this so it's shared

sdks/python/src/lilypad/_internal/otel/types.py
willbakst: Ok I've updated this to explicitly only skip `is None`

sdks/python/src/lilypad/_internal/otel/types.py
willbakst: Ok I wrote a method that should be fairly efficient for doing this properly.

sdks/python/src/lilypad/_internal/otel/_utils.py
```diff
@@ -0,0 +1,81 @@
...
+            or (
+                isinstance(first_item, float)
+                and all(isinstance(item, (int, float)) for item in value)
+            )
+        ):
+            return list(value)
+        elif isinstance(first_item, int | float) and all(
+            isinstance(item, (int, float)) for item in value
+        ):
+            return [float(item) for item in value]
```
koxudaxi, line 71: If `first_item` is a `float`, `int` remains unchanged, but if `first_item` is an `int`, `int` is cast to a `float`, which is as expected.

willbakst: The earlier condition will have already checked the `int` and `float` cases individually, so this just checks it there is a mix of `int | float` and converts the sequence to `list[float]`

koxudaxi:
```or (
                isinstance(first_item, float)
                and all(isinstance(item, (int, float)) for item in value)
            )
```
I apologize for the confusion.
I was talking about a comparison with the previous conditions.

koxudaxi (approved):

sdks/python/src/lilypad/_internal/otel/_utils.py
willbakst: Ah I see, good catch. I've fixed it.

willbakst:
### Merge activity

* **Aug 22, 5:39 PM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/lilypad/680).

sdks/python/src/lilypad/_internal/otel/_utils.py
```diff
@@ -0,0 +1,77 @@
...
+        return True
+    return False
+
+
+def mark_client_as_instrumented(client: Any) -> None:
+    """[MISSING DOCSTRING]"""
+    cast(InstrumentedClient, client).__lilypad_instrumented_client__ = True
+
+
+def serialize_attribute_value(value: Any) -> AttributeValue | None:
```
teamdandelion, line 37: nit: i'd have added unit tests for this function (eg for cases like the mixed float int case)

PR #682, opened 2025-08-22
feat: Add dummy traces endpoint for SDK development

koxudaxi:
### TL;DR

Added OpenTelemetry trace collection endpoint for debugging purposes.

### What changed?

- Replaced Hono with OpenAPIHono for API routing
- Created a new `/v1/traces` endpoint that accepts and logs OpenTelemetry trace data
- Implemented comprehensive Zod schemas for validating OTLP/HTTP trace data
- Added debug logging to display received spans and service information
- Added a new "Telemetry" tag to the OpenAPI documentation

### How to test?

1. Send a POST request to `/v1/traces` with valid OTLP/HTTP formatted trace data
2. Check the logs to verify that span information is being properly received and logged
3. Verify the endpoint returns a valid response with `partialSuccess` object

### Why make this change?

This change enables debugging of OpenTelemetry trace data sent from client applications. By implementing the OTLP/HTTP protocol specification, we can now receive, validate, and log trace data for development and troubleshooting purposes. This is a temporary implementation to help with debugging telemetry data before implementing permanent storage and processing.


koxudaxi:
* **#689** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/689?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#678](https://github.com/Mirascope/lilypad/pull/678) <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/678?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#690](https://github.com/Mirascope/lilypad/pull/690) <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/690?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#682** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/682?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/682?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v1`


<h2></h2>

package.json
```diff
@@ -41,7 +41,7 @@
   },
   "dependencies": {
     "@hono/swagger-ui": "^0.5.2",
-    "@hono/zod-openapi": "^1.1.0",
+    "@hono/zod-openapi": "0.19.10",
```
willbakst: Why change this from major version 1 to 0??

worker/api/traces.ts
willbakst, line 1: The current structure at the top level has a `routes.ts` file with the route definitions, and a `schemas.ts` file with the schema definitions. Here, you have the both the route definition and schema definitions in the same file (`traces.ts`).

Do you think it's better to have everything colocated in the same file like this? For example, we could update the top level to have something like `health.ts` that has the health route and schemas like you have here for traces.

Additionally, you have the implementation for the traces route directly inside of the route. I feel like it will be easier to read things if we stick to a consistent decomposition such as having handlers.

I could imagine having something like `traces.ts` implement a `handleTraces` method that we use in the route (so the route would look like `apiRouter.openapi(tracesRoute, handleTraces);`, which could be nice if we did this consistently for everything).

What do you think? Do you think it would be better to have the schemas, route, and handler all live together in a file (e.g. `traces.ts`) or to have all schemas in `schemas.ts` and all routes in `routes.ts` and handlers in specific files like `traces.ts`?

Of course, we could always break down the latter further into e.g. `schemas/traces.ts` and `routes/traces.ts` and `handlers/traces.ts`. Similarly we could break things down the other way into `traces/routes.ts` and `traces/schemas.ts` and `traces/handlers.ts`.

What would you find easier to organize/understand as the codebase grows? I'm feeling like the `traces/...` structure might be best?

worker/api/traces.ts
```diff
@@ -0,0 +1,127 @@
+import { createRoute, z } from '@hono/zod-openapi';
+
+const KeyValueSchema = z
```
willbakst: Is it common to use "Schema" in the name for schemas? Currently we have things like `HealthResponse` and not `HealthResponseSchema`, so I'm wondering if this should be `KeyValue` instead of `KeyValueSchema`?

This comment is in line with my other comment in that I want to have an open discussion here around how we want to structure and name things so that once we come to a decision we can stick to it for future work.

worker/app.ts
```diff
@@ -43,6 +43,10 @@ app.doc('/openapi.json', {
       name: 'API',
       description: 'Core API endpoints',
     },
+    {
+      name: 'Telemetry',
```
willbakst, line 48: Isn't this technically part of `API`? Why is it separate here?

worker/schemas.ts
```diff
@@ -1,4 +1,4 @@
-import { z } from 'zod';
+import { z } from '@hono/zod-openapi';
```
willbakst: Hmm, why do this indirect import rather than import from `zod` directly? Seems weird?

koxudaxi: If we do not import from this package, the OpenAPI schema JSON will not be generated correctly.

package.json
koxudaxi: We were using zod version 3.x, so we needed to use an older version. We couldn't upgrade zod to version 4.x because different dependencies were not yet compatible.

worker/api/traces.ts
koxudaxi: This follows the naming convention specified in the "@hono/zod-openapi" reference. I'll change it to the naming rule you pointed out.

worker/app.ts
koxudaxi: This is `tags` which is flat structure. I will update the description.

worker/api/traces.ts
koxudaxi: I just remembered. Ultimately,
`export type TraceRequest = z.infer<typeof TraceRequestSchema>;` exports the type, so this naming convention was used.

package.json
willbakst: Why couldn't we use `4.x`? I feel like we should be using the most recent version? This is essentially a new project, so what's preventing that? Also a little confused how this wasn't a problem previously?

worker/api/traces.ts
willbakst: Got it. Let's just make sure we're consistent across the board.

worker/api/traces/schemas.ts
```diff
@@ -0,0 +1,93 @@
...
+    partialSuccess: z
+      .object({
+        rejectedSpans: z.number().optional(),
+        errorMessage: z.string().optional(),
+      })
+      .optional(),
+  })
+  .openapi('TraceResponse');
+
+export type TraceRequest = z.infer<typeof TraceRequestSchema>;
```
willbakst, line 99: What do you think about always exporting the type of every schema?

For example, having `KeyValue` for `KeyValueSchema` etc. for all of them. Essentially we would have a `z.infer` for every single schema?

package.json
koxudaxi: I add "zod": "4.x" in dependencies. It resolve the problem.

worker/api/traces/schemas.ts
koxudaxi: I exported all schemas.

willbakst (approved):

willbakst:
### Merge activity

* **Sep 10, 3:37 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/lilypad/682).

PR #681, opened 2025-08-21
feat: add support for .stream methods that use context managers

willbakst:
Add support for the `.stream` methods for OpenAI and Anthropic, which use context managers.

Note that async context managers are not awaitable, so we instrument those methods using the sync path. I've added notes to that effect in the code so it's clear.

PR #686, opened 2025-08-22
chore: make public vs. private naming consistent

willbakst:


willbakst:
* **#686** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/686?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/686?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#685** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/685?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#684** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/684?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#681** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/681?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#680** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/680?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v1`


<h2></h2>

sdks/python/src/lilypad/_internal/otel/anthropic.py
```diff
@@ -297,13 +299,13 @@ def _process_chunk(
 
 
 def instrument_anthropic(
-    client: AnthropicClient,
+    client: _AnthropicClient,
```
koxudaxi: Isn't it better to use public rather than private for the argument type of public functions?

sdks/python/src/lilypad/_internal/otel/openai.py
```diff
@@ -274,7 +274,7 @@ def _process_chunk(
 
 
 def instrument_openai(
-    client: OpenAIClient,
+    client: _OpenAIClient,
```
koxudaxi: Isn't it better to use public rather than private for the argument type of public functions?

koxudaxi (commented):
I left two comments.

sdks/python/src/lilypad/_internal/otel/anthropic.py
willbakst: Good call. I've updated

sdks/python/src/lilypad/_internal/otel/openai.py
willbakst: Good call. I've updated

koxudaxi (approved):

sdks/python/src/lilypad/_internal/otel/_utils.py
```diff
@@ -31,7 +31,7 @@ def client_is_already_instrumented(client: Any) -> bool:
 
 def mark_client_as_instrumented(client: Any) -> None:
```
teamdandelion, line 32: Wondering if this should be private too? Presumably this should get called when the client is actually instrumented. Calling it without actually instrumenting hte client seems like a setup for confusion. 

(Maybe it's implicitly private by virtue of being in _utils, but since you renamed InstrumentedClient, I infer there isn't a general expectation that everything in this file is implicitly private.)

PR #683, opened 2025-08-22
test: Add unit tests for traces endpoint

koxudaxi:
### TL;DR

Added tests for the OpenTelemetry traces endpoint.

### What changed?

Added comprehensive test coverage for the `/api/v1/traces` endpoint that handles OpenTelemetry trace data. The tests verify:

- Successful processing of valid OTLP trace data
- Handling of multiple spans and resource spans
- Processing traces without service name
- Handling empty resource spans
- Proper response structure with partial success fields
- Rejection of invalid trace data
- Error handling for malformed JSON
- Extraction of service name from resource attributes
- Processing of complex attribute values
- OpenAPI documentation for the traces endpoint

### How to test?

Run the test suite with:

```bash
npm test worker/api/traces.test.ts
```

The tests use Vitest and mock console.log to verify the correct logging of trace information.

### Why make this change?

To ensure the OpenTelemetry traces endpoint functions correctly and maintains expected behavior through future changes. These tests validate both the successful processing of valid trace data and proper error handling for edge cases, providing confidence in the reliability of the telemetry collection system.

PR #687, opened 2025-08-22
feat: Generate and add Python SDK client using Fern

koxudaxi:
### TL;DR

Added SDK generation infrastructure using Fern to create client libraries from our OpenAPI specification.

### What changed?

- Added Fern API SDK generator configuration in `sdks/fern/` directory
- Created initial OpenAPI specification in `sdks/fern/openapi.json`
- Generated Python SDK in `sdks/python/src/lilypad/_generated/`
- Added npm scripts for SDK generation and validation
- Updated file structure documentation to include the new SDK section
- Added server name identifiers to OpenAPI specification for Fern compatibility
- Added Fern API dependency to the project

### How to test?

1. Run `bun run sdk:extract-openapi` to extract the current OpenAPI spec
2. Run `bun run sdk:generate` to generate the Python SDK
3. Verify the generated SDK files in `sdks/python/src/lilypad/_generated/`
4. Run `bun run sdk:validate` to ensure the OpenAPI spec is up-to-date

### Why make this change?

This change establishes the foundation for automatically generating client SDKs from our API specification. By using Fern to generate SDKs, we can:

1. Provide consistent, well-documented client libraries for multiple languages
2. Ensure SDKs stay in sync with our API as it evolves
3. Reduce manual work maintaining client libraries
4. Improve developer experience for API consumers

The initial implementation focuses on Python, but the infrastructure is in place to support additional languages in the future.


koxudaxi:
* **#689** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/689?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#678](https://github.com/Mirascope/lilypad/pull/678) <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/678?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#690](https://github.com/Mirascope/lilypad/pull/690) <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/690?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#687** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/687?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/687?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v1`


<h2></h2>

docs/file-structure.md
```diff
@@ -66,7 +67,9 @@ The serverless backend running on Cloudflare Workers that handles API requests a
 worker/
 ├── api/                  # API handlers
 │   ├── index.ts            # API route handlers
-│   └── router.ts           # API router configuration
+│   ├── router.ts           # API router configuration
+│   ├── traces.ts           # OpenTelemetry traces endpoint
```
willbakst: nit: this may change depending on down stack, so noting here (namely the additional decomposition into `traces/...`

docs/file-structure.md
```diff
@@ -148,6 +151,34 @@ db/
 - **Migration system**: Version-controlled database schema changes
 - **Vitest**: Test runner with comprehensive test coverage and isolated test environment
 
+## Client SDKs (`sdks/`)
+
+Auto-generated client SDKs for interacting with the Lilypad API.
```
willbakst: nit: technically only the `_generated` folders are auto-generated since we are manually writing everything else in the SDKs

package.json
```diff
@@ -36,6 +36,10 @@
     "lint:ci": "bun run codespell; bun run lint; bun run lint:python-sdk",
     "fix": "bun run format; bun run lint:eslint:fix",
     "fix:python-sdk": "cd sdks/python && uv run ruff check --fix . && uv run ruff format .",
+    "sdk:extract-openapi": "curl -s http://localhost:3000/openapi.json | npx prettier --parser json > sdks/fern/openapi.json",
```
willbakst: I think we can do a better job naming these scripts.

For example, I think `sdks:openapi:extract` would be better for this command. Also, this command assumes the server is running. I wonder if we should spin up the server, extract the OpenAPI, and then spin down the server all as part of this so that we can more reliably run it in husky pre-commit?

package.json
```diff
@@ -36,6 +36,10 @@
...
+    "sdk:generate": "cd sdks/fern && npx fern generate --group python-sdk --local --force && bun run fix:python-sdk",
```
willbakst: I would update this to `sdks:python:generate`? Also, what do you think about having this script also extract the current OpenAPI JSON first? I can't really imagine where I would want to run this without first making sure the OpenAPI JSON is up to date?

package.json
```diff
@@ -36,6 +36,10 @@
...
+    "sdk:validate": "bun run sdk:extract-openapi && git diff --exit-code sdks/fern/openapi.json",
```
willbakst: similarly I would update this to `sdks:openapi:validate`?

package.json
```diff
@@ -36,6 +36,10 @@
...
+    "sdk:generate-all": "bun run sdk:extract-openapi && bun run sdk:generate",
```
willbakst: In line with my other comment, `sdks:python:generate` would cover this, so this command would be unnecessary and could be deleted?

sdks/fern/fern.config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "organization": "lilypad-sdk",
```
willbakst: I believe we have to make our organization `"mirascope"`

I see in the README below that the `pip install lilypad-sdk` is correct, but I'm wondering if we could ensure that with a separate field since organization has to be "mirascope" since that's our organization name with Fern.

sdks/fern/generators.yml
```diff
@@ -0,0 +1,19 @@
...
+    generators:
+      - name: fernapi/fern-python-sdk
+        version: 4.21.2
+        output:
+          location: local-file-system
+          path: ../python/src/lilypad/_generated
+        smart-casing: true
+        config:
+          client_class_name: Lilypad
+          package_name: lilypad._generated
```
willbakst, line 18: What about `filename: client.py`? Looks like that's the filename anyway, but I remember having this setting in the previous yml?

sdks/fern/generators.yml
```diff
@@ -0,0 +1,19 @@
...
+          pydantic_config:
+            use_str_enums: false
```
willbakst: Why do we have this setting? Is there a benefit to having this as `false` instead of `true`?

sdks/fern/openapi-overrides.yml
willbakst, line 1: Can you explain the design decision here around naming?

sdks/python/pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "opentelemetry-instrumentation>=0.57b0",
     "opentelemetry-sdk>=1.36.0",
     "orjson>=3.10.0",
+    "pydantic>=2.11.7",
```
willbakst: Is this the minimum necessary version or the most recent? I feel like best practice would be to make the required version as low as possible here?

sdks/python/pyproject.toml
```diff
@@ -86,6 +87,8 @@ target-version = "py310"
 "__init__.py" = ["F401"]
 "tests/**.py" = ["T201", "T203", "ANN", "UP007"]
 "examples/**.py" = ["T201", "T203"]
+"src/lilypad/_generated/**/*.py" = ["ANN", "D", "E501", "UP045"]
```
willbakst: Since this is generated by Fern, I wonder if we should just ignore everything for it? Is there a benefit to ignoring only these specific rules?

sdks/python/pyproject.toml
```diff
@@ -154,6 +157,7 @@ pythonVersion = "3.10"
 venvPath = "."
 venv = ".venv"
 include = ["src", "tests", "examples"]
+ignore = ["src/lilypad/_generated/**"]
```
willbakst: Do you need the `**` here or would `src/lilypad/_generated` work?

docs/file-structure.md
```diff
@@ -148,6 +151,34 @@ db/
...
+│   ├── fern.config.json      # Fern CLI configuration
+│   ├── generators.yml        # SDK generator configuration
+│   ├── openapi.json          # OpenAPI specification (generated)
+│   └── openapi-overrides.yml # SDK method name overrides
+└─── python/                # Python SDK
+    ├── src/
+    │   └── lilypad/
+    │       ├── _generated/   # Generated SDK code
+    │       ├── _internal/    # Internal instrumentation code
+    │       ├── api_client.py # Public API wrapper
```
willbakst: What is this file? I don't see it anywhere in this PR? Same for `client.py` below?

sdks/python/src/lilypad/_generated/__init__.py
willbakst, line 1: note: we will need to think through our process for re-exporting these at the top level so users can do something like `from lilypad import Lilypad`. Not sure if a `*` import would be best so everything is always exported or if we should be mindful in each PR of what we're choosing to export. The latter is likely safer

docs/file-structure.md
koxudaxi: I just updated it

sdks/fern/fern.config.json
koxudaxi: https://buildwithfern.com/learn/docs/getting-started/project-structure#fern-definition

Right. Changing this here will alter the package name. I don't think there are any additional fields.

sdks/fern/generators.yml
koxudaxi: OK, I added the attributes

sdks/fern/generators.yml
koxudaxi: I see I had a leftover from my various attempt, so I'll remove them.

sdks/fern/openapi-overrides.yml
koxudaxi: I changed the naming to be more purpose-driven, such as `send_traces`. Would you prefer the simpler naming convention used previously, like `create`?

sdks/python/pyproject.toml
koxudaxi: make sense. ruff ignore all rules

sdks/python/src/lilypad/_generated/__init__.py
koxudaxi: I will add it in each pull request.

sdks/python/pyproject.toml
koxudaxi: I just fixed it.

sdks/fern/openapi-overrides.yml
willbakst: Hmm, I'm not sure. It's not that I dislike `client.telemetry.send_traces`, but I wonder if that would be more confusing compared to `client.traces.create` (i.e. what would be more clear to a user given existing conventions)? As we add more endpoints, will we do things like `client.get_spans` and `client.update_span` or will we do something more like `client.spans.get` and `client.spans.update`?

Similarly, for "health" I wonder if `client.check` is clear or if it should be `client.health.check`? Or `client.health`?

sdks/python/pyproject.toml
```diff
@@ -86,6 +88,7 @@ target-version = "py310"
 "__init__.py" = ["F401"]
 "tests/**.py" = ["T201", "T203", "ANN", "UP007"]
 "examples/**.py" = ["T201", "T203"]
+"src/lilypad/api_client.py" = ["ANN201"]
```
willbakst: What is this? There isn't even a file with this name?

sdks/fern/fern.config.json
willbakst: Isn't the `package_name` field in `generators.yml` what determines that? Also we will write our own release scripts with `uv` where we will specify all of this, no?

sdks/python/pyproject.toml
koxudaxi: Oh, I forgot to remove it.

sdks/fern/fern.config.json
koxudaxi: Sorry for the insufficient explanation. It's not the actual package, but the package_name used in the generated README.md and other generated documents. I thought organization was a value for defining these.
Should I set this to mirascope and use sed or something to replace mirascope with lilypad-sdk in the generated documents? Or should I delete the documents entirely?

sdks/fern/openapi-overrides.yml
koxudaxi: I agree with you. I'll switch to client.traces.create and client.health.check. much more consistent with existing conventions. I will update the PR.

koxudaxi: I fixed it.

sdks/fern/fern.config.json
willbakst: I've updated this to have `"organization": "mirascope"`, which is correct, and I've reached out to their team to see if we can work to resolve this issue somehow. We can revisit later. For now, not a big deal that the README is incorrect since that import won't work anyway since we haven't released this version yet.

willbakst (approved):

willbakst:
### Merge activity

* **Sep 10, 3:53 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/lilypad/687).

PR #688, opened 2025-08-22
chore: Add validation scripts for generated OpenAPI and SDK

koxudaxi:

PR #689, opened 2025-08-26
feat: Add Lilypad Clients

koxudaxi:


koxudaxi:
* **#689** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/689?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#678](https://github.com/Mirascope/lilypad/pull/678) <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/678?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#690](https://github.com/Mirascope/lilypad/pull/690) <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/690?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/689?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v1`


<h2></h2>

docs/file-structure.md
```diff
@@ -165,10 +165,17 @@ sdks/
 └─── python/                # Python SDK
     ├── src/
     │   └── lilypad/
-    │       ├── _generated/   # Generated SDK code
-    │       ├── _internal/    # Internal instrumentation code
+    │       ├── _generated/   # Generated SDK code (DO NOT MODIFY)
```
willbakst: Since this is generated, I don't think it's worth documenting here deeper than just `_generated` and noting it's generated by Fern.

Also, when you say `(DO NOT MODIFY)`, it seems like it might be better to say `(DO NOT MODIFY MANUALLY)` since it would technically be ok to modify it so long as it's through the auto-generation process?

docs/file-structure.md
```diff
@@ -165,10 +165,17 @@ sdks/
...
+    │       │   ├── client.py    # Fern-generated base client
+    │       │   ├── errors/      # API error classes
+    │       │   └── telemetry/   # Telemetry API clients
+    │       ├── _internal/    # Internal utilities
+    │       │   ├── errors.py    # Custom error classes
+    │       │   ├── otel/        # OpenTelemetry instrumentation
```
willbakst: Ah, I should've updated this in my PR. Why not add the deeper structure here with this change? Why stop at `otel/`?

docs/file-structure.md
```diff
@@ -165,10 +165,17 @@ sdks/
...
+    │       │   └── settings.py  # SDK settings management
     │       ├── api_client.py # Public API wrapper
```
willbakst: again, not sure what this file is?

docs/file-structure.md
```diff
@@ -165,10 +165,17 @@ sdks/
...
-    │       └── client.py     # Main client module
+    │       ├── client.py     # Enhanced client with caching
+    │       └── configuration.py # SDK configuration utilities
```
willbakst: nit: weird the comments don't align here? let's make sure they are left aligned with each other please. This may require further indenting everything to keep the proper nest level correct across the board

sdks/python/pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
     "opentelemetry-sdk>=1.36.0",
     "orjson>=3.10.0",
     "pydantic>=2.11.7",
+    "pydantic-settings>=2.10.1",
```
willbakst: nit: same thing here re: making sure we fix this to the minimum possible version for maximal compatibility with user code?

sdks/python/src/lilypad/_internal/settings.py
```diff
@@ -0,0 +1,50 @@
...
+from pydantic import Field
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+
+class Settings(BaseSettings):
+    """Global settings for Lilypad SDK."""
+
+    base_url: str | None = Field(default=None)
+    api_key: str | None = None
+    project_id: str | None = None
```
willbakst: I think going forward since `api_key` is bound to a specific project, it will be cleaner not to require the user provide the project ID? In `v0` we have `/projects/{:projectId}` everywhere, and I feel like that nesting is a bit annoying? WDYT?

sdks/python/src/lilypad/_internal/settings.py
```diff
@@ -0,0 +1,50 @@
...
+def get_settings() -> Settings:
+    """Return Settings for the current context."""
+    settings = _current_settings.get()
+    if settings is None:
+        settings = _default_settings()
+        _current_settings.set(settings)
+    return settings
+
+
+def _set_settings(settings: Settings) -> None:
```
willbakst: Looks like we don't actually use this anywhere? I would remove

sdks/python/src/lilypad/_internal/settings.py
```diff
@@ -0,0 +1,50 @@
...
+def _default_settings() -> Settings:
+    return Settings()
+
+
+_current_settings: ContextVar[Settings | None] = ContextVar(
+    "_current_settings", default=None
+)
+
+
+def get_settings() -> Settings:
```
willbakst, line 39: Hmm, this method feels a bit weird in that it updates the context but is not a context manager? Why not make this a context manager that takes arguments for each setting? This would make it possible for a user to easily override specific settings at runtime?

Or is this the getter for getting the context? In that case, where is the context manager? We should have that right? A `def settings(...) -> Settings: ...` method marked with `@contextmanager`?

sdks/python/src/lilypad/client.py
```diff
@@ -0,0 +1,267 @@
...
+            )
+
+        except Exception as e:
+            logger.error("Failed to initialize AsyncLilypad client: %s", e)
+            raise RuntimeError(f"Async client initialization failed: {e}") from e
+
+
+@lru_cache(maxsize=256)
+def _sync_singleton(api_key: str, base_url: str | None) -> Lilypad:
+    """Return (or create) the process-wide synchronous client."""
```
willbakst: nit: update to "Return the process-wide synchronous client, creating one if none yet exists"

sdks/python/src/lilypad/_internal/settings.py
```diff
@@ -0,0 +1,50 @@
...
+
+    model_config = SettingsConfigDict(env_prefix="LILYPAD_")
+
+
+@cache
+def _default_settings() -> Settings:
+    return Settings()
+
+
+_current_settings: ContextVar[Settings | None] = ContextVar(
```
willbakst: nit: make this `CURRENT_SETTINGS`?

sdks/python/src/lilypad/client.py
```diff
@@ -0,0 +1,267 @@
...
+        base_url: str | None = None,
+        api_key: str | None = None,
+        token: Token = None,
+        timeout: float | None = None,
+        follow_redirects: bool | None = True,
+        httpx_client: httpx.Client | None = None,
+    ) -> None:
+        """Initialize the enhanced Lilypad client."""
+        try:
+            effective_api_key = api_key or get_settings().api_key
```
willbakst: nit: why not just do a single `settings = get_settings()` at the top here? You do a bunch of `get_settings().{field}` access, seems better to just get the settings at the beginning?

willbakst: nit: why not just set the `self.api_key` variable here? And then do `if not self.api_key` below?

sdks/python/src/lilypad/client.py
```diff
@@ -0,0 +1,267 @@
...
+                httpx_client = httpx.Client(
+                    headers=headers,
+                    timeout=timeout or 30.0,
+                    follow_redirects=follow_redirects
+                    if follow_redirects is not None
+                    else True,
+                )
+
+            super().__init__(
+                base_url=base_url or get_settings().base_url,
```
willbakst: Why not just set `base_url` separately here and then below do `self.base_url = base_url or self._client_wrapper.get_base_url()`? Also, feels wrong to be using `self._client_wrapper` given it's private? Seems like it would be better for us to set the correct default `base_url` at the settings level if not provided by the user? Or is that annoying because it could depend on the environment?

sdks/python/src/lilypad/client.py
```diff
@@ -0,0 +1,267 @@
...
+                or get_settings().base_url
+                or self._client_wrapper.get_base_url()
+            )
+
+        except Exception as e:
+            logger.error("Failed to initialize Lilypad client: %s", e)
+            raise RuntimeError(f"Client initialization failed: {e}") from e
+
+
+class AsyncLilypad(_BaseAsyncLilypad):
```
willbakst, line 83: All the same comments above for the sync client apply here

sdks/python/src/lilypad/client.py
```diff
@@ -0,0 +1,267 @@
...
+    """Get or create a cached synchronous client.
+
+    Args:
+        api_key: API key for authentication
+        base_url: Base URL for the API
+
+    Returns:
+        Cached Lilypad client instance
+    """
+    key = api_key or get_settings().api_key
```
willbakst: why not just overwrite `api_key` and `base_url`? It's not like you need to preserve the original values if they were `None`, right?

sdks/python/src/lilypad/client.py
```diff
@@ -0,0 +1,267 @@
...
+        Cached AsyncLilypad client instance
+    """
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError as exc:
+        raise RuntimeError(
+            "get_async_client() must be called from within an active event loop."
+        ) from exc
+
+    key = api_key or get_settings().api_key
```
willbakst: same thing here

sdks/python/src/lilypad/client.py
```diff
@@ -0,0 +1,267 @@
...
+
+    effective_base_url = base_url or get_settings().base_url
+    logger.debug(
+        "Creating async client with api_key=*****, base_url=%s", effective_base_url
+    )
+
+    return _async_singleton(key, id(loop), effective_base_url)
+
+
+def create_transport_client(
```
willbakst, line 218: What is the different between this and the sync client creation above? Is it the increased default timeout?

sdks/python/src/lilypad/client.py
```diff
@@ -0,0 +1,267 @@
...
+    )
+
+
+def close_cached_clients() -> None:
+    """Close all cached client instances."""
+    _sync_singleton.cache_clear()
+    _async_singleton.cache_clear()
+
+
+__all__ = [
```
willbakst: How many of these methods do we actually expect users to use? I wonder if we should put this inside of our internal module as `_internal/client.py` and then only export what is necessary for users such as `Lilypad` and `AsyncLilypad`?

sdks/python/src/lilypad/__init__.py
```diff
@@ -4,11 +4,30 @@
 
 from . import configuration
 from .configuration import configure
+from .client import (
```
willbakst: In line with my other comment, we should be mindful of what we keep internal vs. export for users.

sdks/python/pyproject.toml
koxudaxi: Like pydantic, I have set the minimum version to support Python 3.13.
Supporting every Python version would increase maintenance costs too much, so I will base it on Python 3.13.

sdks/python/src/lilypad/_internal/settings.py
koxudaxi: Great. Now that you mention it, you're right. I'll delete it immediately.

sdks/python/src/lilypad/client.py
koxudaxi: The differences are:

1. Caching: `get_sync_client()` returns a cached singleton via `@lru_cache`, while `create_transport_client()` creates a new instance each time
2. Settings fallback: `get_sync_client()` uses `get_settings()` for defaults, `create_transport_client()` doesn't
3. Extra params: `create_transport_client()` accepts timeout and custom httpx_client

`create_transport_client()` is for OpenTelemetry transport where we need fresh instances with custom configs. `get_sync_client()` is for general use with client reuse.

sdks/python/src/lilypad/client.py
koxudaxi: This is unnecessary at this stage of the PR, so I will delete it.

docs/file-structure.md
```diff
@@ -157,7 +157,7 @@ db/
 
 ## Client SDKs (`sdks/`)
 
-Client SDKs for interacting with the Lilypad API, combining auto-generated and manually written code.
+Auto-generated client SDKs for interacting with the Lilypad API.
```
willbakst: This change is incorrect and against my previous comment? The previous language was correct.

docs/file-structure.md
```diff
@@ -166,16 +166,25 @@ sdks/
 │   ├── generators.yml        # SDK generator configuration
 │   ├── openapi.json          # OpenAPI specification (generated)
 │   └── openapi-overrides.yml # SDK method name overrides
-└─── python/                # Python SDK
+└─── python/                         # Python SDK
```
willbakst: This comment should be at the same level as fern since it's the same level in the file system. You should further indent all comments to give them enough space so they are minimum two spaces further than the right-most character of the file diagram (and then left aligned to match indentation of comments with the file diagram).

If this isn't clear let me know and I will explain in more detail.

docs/file-structure.md
```diff
@@ -166,16 +166,25 @@ sdks/
...
-    │       └── configuration.py # SDK configuration
-    ├── examples/           # Example usage scripts
-    └── tests/              # SDK tests
+    │       ├── _generated/          # Generated by Fern (DO NOT MODIFY MANUALLY)
+    │       ├── _internal/           # Internal utilities
+    │       │   ├── errors.py        # Custom error classes
+    │       │   ├── otel/            # OpenTelemetry instrumentation
+    │       │   │   ├── __init__.py           # OTel exports
+    │       │   │   ├── _utils.py             # OTel utilities
+    │       │   │   ├── anthropic.py          # Anthropic instrumentation
```
willbakst: nit: should be `Anthropic instrumentation` like you have for OpenAI below.

docs/file-structure.md
```diff
@@ -166,16 +166,25 @@ sdks/
...
+    │       │   │   ├── base_instrumentor.py  # Base instrumentor class
+    │       │   │   ├── openai.py             # OpenAI instrumentation
+    │       │   │   ├── stream_wrappers.py    # Stream wrapper utilities
```
willbakst: These comments aren't doing much besides putting the file name in normal language formatting. It would be better to do something like "Shared utilities for instrumenting streams"? Same for "Base instrumentor class" which would likely be better as something like "Base class for shared LLM OTel instrumentation implementation details"?

sdks/python/pyproject.toml
willbakst: That makes sense. I think the minimum version that supports 3.13 is a good baseline. Ideally we will continue to support the minimum possible version, but that shouldn't come as a maintenance burden cost. Better to bump versions up in that case to make maintenance easier (if this happens in the future).

sdks/python/src/lilypad/_internal/settings.py
```diff
@@ -0,0 +1,80 @@
...
+from contextlib import contextmanager
+
+from pydantic import Field
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+
+class Settings(BaseSettings):
+    """Global settings for Lilypad SDK."""
+
+    base_url: str = Field(default="https://app.lilypad.so")
```
willbakst: nit: default should be `https://v1.lilypad.mirascope.com` with the new domain update, right? Eventually we'll drop the `v1` when we release

sdks/python/src/lilypad/_internal/settings.py
```diff
@@ -0,0 +1,80 @@
...
+        with settings(base_url="https://api.example.com", api_key="test-key"):
+            # Use custom settings within this context
+            client = LilypadClient()
+    """
+    current = CURRENT_SETTINGS.get()
+    if current is None:
+        current = _default_settings()
+
+    new_settings = Settings(
+        base_url=base_url if base_url is not None else current.base_url,
```
willbakst: why not just `base_url=base_url or current.base_url`? Same for `api_key` below?

sdks/python/src/lilypad/client.py
```diff
@@ -0,0 +1,247 @@
...
+    logger.debug("Creating sync client with api_key=*****, base_url=%s", base_url)
+
+    return _sync_singleton(api_key, base_url)
+
+
+@lru_cache(maxsize=256)
+def _async_singleton(
+    api_key: str, loop_id_for_cache: int, base_url: str | None = None
+) -> AsyncLilypad:
+    """Return (or create) an asynchronous client bound to a specific loop."""
```
willbakst: nit: update this comment to match the sync comment update we made

docs/file-structure.md
koxudaxi: I reverted it

docs/file-structure.md
koxudaxi: I can see `# Anthropic instrumentation` 
Is the formatting already consistent?

willbakst: I missed it from the wrap, that's my bad.

docs/file-structure.md
willbakst: I've updated this and submitted the change. Please see the update as a reference for the future.

willbakst (approved):

PR #691, opened 2025-08-27
test: Add Lilypad exporter unittest

koxudaxi:
### TL;DR

Simplified OpenTelemetry exporters code and added comprehensive test suite for telemetry functionality.

### What changed?

- Removed verbose documentation comments from exporter classes
- Simplified error handling in the transport layer
- Reduced thread pool size in LLMSpanProcessor from 2 to 1
- Added extensive test suite with VCR cassettes for HTTP interaction testing
- Added unit tests for all exporter components (processors, transport, config)
- Added integration tests that verify the full telemetry pipeline

### How to test?

Run the new test suite:
```bash
pytest sdks/python/tests/lilypad/_internal/otel/exporters/
```

The tests use VCR cassettes to simulate HTTP interactions, so they should pass without requiring a real server connection.

### Why make this change?

This change improves the maintainability of the OpenTelemetry exporters by:
1. Removing redundant documentation that duplicated the class/method signatures
2. Simplifying the code to focus on core functionality
3. Adding comprehensive test coverage to ensure reliability
4. Creating VCR cassettes that document the expected HTTP interactions for future reference

The test suite provides examples of how to use the exporters in various scenarios, making it easier for developers to understand the telemetry system.

PR #690, opened 2025-08-26
feat: Add Lilypad Client implementation

koxudaxi:

PR #679, opened 2025-08-21
Implement Otel Exporters details without a client

koxudaxi:
### TL;DR

Implemented the OpenTelemetry exporters for Lilypad, enabling trace data collection and transmission to the Lilypad backend.

### What changed?

- Implemented the `configure_exporters` function to set up the OpenTelemetry exporters
- Added implementation for `ImmediateStartExporter` and `LilypadOTLPExporter` classes
- Implemented the `_create_start_event` method in `LLMSpanProcessor`
- Refactored the transport layer to use a client-based approach that will work with the future Fern-generated client
- Renamed types for better clarity (`StatusDict` → `Status`, `LinkDict` → `Link`)
- Added utility functions for formatting trace and span IDs
- Updated the transport implementation to use a configuration-based approach
- Implemented span conversion logic to transform OpenTelemetry spans into Lilypad's format

### How to test?

1. Initialize the Lilypad SDK with a valid API key and endpoint
2. Create spans using the OpenTelemetry API
3. Verify that span start events are sent immediately
4. Verify that completed spans are batched and sent to the Lilypad backend
5. Check that all span attributes, events, and links are properly transmitted

Note: The implementation currently raises a `NotImplementedError` when initializing the transport client without a provided Fern client, as the Fern-generated client is not yet available.

### Why make this change?

This change enables the Lilypad SDK to collect and transmit OpenTelemetry trace data to the Lilypad backend, which is essential for monitoring and analyzing LLM operations. The implementation follows OpenTelemetry best practices while providing Lilypad-specific functionality like immediate start event transmission for real-time visibility into LLM operations.

PR #692, opened 2025-08-29
feat: Add trace/span interface

koxudaxi:
### TL;DR

Added core tracing, versioning, and remote execution capabilities to the Python SDK with a clean API design.

### What changed?

This PR introduces three key decorator-based capabilities to the Lilypad Python SDK:

1. **Tracing** - Added `@trace` decorator and `span()` function for explicit tracing with detailed span management
2. **Versioning** - Added `@version` decorator for function versioning with ability to run specific versions
3. **Remote Execution** - Added `@remote` decorator for executing functions on remote targets

The implementation includes:
- Type definitions with Protocol classes for better IDE support
- Custom exception hierarchy with `LilypadException` as the base class
- `TraceResult` class for per-call operations like annotation, tagging, and assignment
- Proper exports in `__init__.py` for public API access

Most implementation methods are marked with `NotImplementedError` as this PR focuses on establishing the API design first.

### How to test?

The implementation is not yet functional as most methods raise `NotImplementedError`. Once implemented, you could test with:

```python
from lilypad import trace, span, remote, version

# Test tracing
@trace
def my_function(x):
    with span("processing"):
        return x * 2

result = my_function(5)  # Normal execution
wrapped = my_function.wrap(5)  # Traced execution
wrapped.annotate(label="Example", reasoning="Testing the API")

# Test remote execution
@remote
def remote_function(data):
    return data.upper()

remote_result = await remote_function.execute_remote("gpu-worker", "test")

# Test versioning
@version(tag="v1.0")
def versioned_function():
    return "current version"

old_result = versioned_function.run_version("v0.9")
```

### Why make this change?

This change establishes a foundation for advanced capabilities in the Lilypad SDK:

1. The tracing system will allow users to monitor, debug, and analyze function execution with detailed context
2. Versioning enables reproducibility and controlled evolution of functions
3. Remote execution provides a clean way to offload computation to specialized workers

The design prioritizes a clean, composable API that follows Python idioms while providing strong type hints for IDE support. The decorators can be used individually or composed together to create functions with multiple capabilities.


koxudaxi:
* **#694** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/694?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#695](https://github.com/Mirascope/lilypad/pull/695) <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/695?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#696](https://github.com/Mirascope/lilypad/pull/696) <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/696?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#693** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/693?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#692** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/692?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/692?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#691** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/691?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#679** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/679?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#678** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/678?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#689** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/689?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#690](https://github.com/Mirascope/lilypad/pull/690) <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/690?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `v1`


<h2></h2>

willbakst:
I wonder if the `@version` decorator should implement a `VersionedFunctionDecorator` class that implements the decorator, which would then return a `VersionedFunction` class. This class would then provide all of the utility methods that things like e.g. `mode="wrap"` provided in `v0` without having to set the mode explicitly. We would then also implement the `__call__` method so that calling the versioned function would operate exactly as though the user called the original function. A benefit of having the decorator return a class is that we can more easily put methods like `.version` or `.remote` on the class like we do for `.stream` and `.resume` for the e.g. `Call` class in Mirascope `v2`.

One thing to note is that we should be mindful of preserving function information as though we used `functools.wraps` to wrap the function so that doing things on the returned class like checking its name would return the same results as though you did it on the original function.

sdks/python/src/lilypad/remote/__init__.py
willbakst, line 1: As discussed, we can completely ignore remote execution for now. In the future, we may add it as a simple wrapper on the `.version` utility method.

sdks/python/src/lilypad/spans.py
```diff
@@ -0,0 +1,247 @@
...
+
+from __future__ import annotations
+
+from typing import Any, Protocol, runtime_checkable
+
+from opentelemetry.trace import SpanContext
+
+
+@runtime_checkable
+class SpanProtocol(Protocol):
```
willbakst: What is the point of the protocol vs. just implementing `Span`?

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,339 @@
...
+    overload,
+    Protocol,
+)
+from typing_extensions import TypeIs
+
+P = ParamSpec("P")
+R = TypeVar("R", covariant=True)
+
+# Type for annotation data - supports JSON-serializable values
+JSONValue = Union[
```
willbakst: nit: in Mirascope we name this `Jsonable` and implement it as such:

```python
Jsonable: TypeAlias = (
    None
    | str
    | int
    | float
    | bool
    | Sequence["Jsonable"]
    | Mapping[str, "Jsonable"]
    | JsonableObject
)
"""Simple type alias for JSON-serializable types."""
```

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,339 @@
...
+            The function UUID or None if not available.
+        """
+        return self._function_uuid
+
+
+@dataclass(kw_only=True)
+class TraceDecorator:
+    """[MISSING DOCSTRING]"""
+
+    config: _TraceConfig = field(default_factory=_TraceConfig)
```
willbakst: Again, not really sure what the benefit of this additional indirection with the configuration is?

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,339 @@
...
+    Creates a wrapper that enables distributed tracing, performance monitoring,
+    and execution tracking for decorated functions.
+
+    Args:
+        tags: Optional list of string tags to associate with traced executions.
+
+    Returns:
+        A decorator that wraps functions with tracing capabilities.
+
+    Examples:
```
willbakst: Google-style docstrings would have a singular `Example: ...` section and the code would be in python code blocks. Multiple examples would be included by having multiple of the singular `Example: ...` sections

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,339 @@
...
+        >>> @trace()
+        ... def process_data(data: dict) -> dict:
+        ...     return {"processed": data}
+
+        >>> @trace(tags=["production", "critical"])
+        ... async def fetch_user(user_id: int) -> User:
+        ...     return await get_user(user_id)
+
+    Raises:
+        NotImplementedError: This function is not yet implemented.
```
willbakst: seems implemented to me?

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,339 @@
...
+class AsyncFunction(Protocol[P, R]):
+    """[MISSING DOCSTRING]"""
+
+    async def __call__(self, *args: P.args, **kwargs: P.kwargs) -> R:
+        """[MISSING DOCSTRING]"""
+        ...
+
+
+@dataclass(kw_only=True)
+class TracedFunctionBase(ABC):
```
willbakst: nit: should be `BaseTracedFunction`? Also a little weird that it inherits from `ABC` but there's nothing abstract about it

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,339 @@
...
+    config: _TraceConfig
+    _is_async: bool
+    _qualified_name: str = field(init=False)
+    _module_name: str = field(init=False)
+
+
+@dataclass(kw_only=True)
+class TracedFunction(Generic[P, R], TracedFunctionBase):
```
willbakst: It seems like `TracedFunction` here and `AsyncTracedFunction` below are missing their interfaces for things like `__call__` and other possible utility methods (e.g. a method that would return `TraceResult`?)

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,339 @@
...
+
+
+@dataclass
+class _TraceConfig:
+    """Configuration for trace decorator."""
+
+    tags: tuple[str, ...] = ()
+
+
+class AsyncTraceResult(Generic[R]):
```
willbakst: This seems the same as `TraceResult`? Shouldn't it be async on certain methods like `annotate` etc.?

sdks/python/src/lilypad/types.py
```diff
@@ -0,0 +1,178 @@
...
+from typing import Any, Awaitable, Generic, ParamSpec, Protocol, TypeVar
+
+from .versioning.types import VersionId, VersionRef
+
+P = ParamSpec("P")
+R = TypeVar("R", covariant=True)
+T = TypeVar("T")
+
+
+class TraceResult(Generic[R]):
```
willbakst: This is duplicate? Also duplicated with `Traced` below? Please dedupe

sdks/python/src/lilypad/types.py
```diff
@@ -0,0 +1,178 @@
...
+    def tag(self, *tags: str) -> None:
+        """Global tagging helper (no-op by default, prefer per-call via wrap)."""
+        ...
+
+    def assign(self, *emails: str) -> None:
+        """Global assignment helper (no-op by default, prefer per-call via wrap)."""
+        ...
+
+
+class Versioned(Traced[P, R], Protocol[P, R]):
```
willbakst: Hmm, shouldn't this be in the versioning stuff? What is it doing here?

sdks/python/src/lilypad/versioning/decorator.py
```diff
@@ -0,0 +1,166 @@
...
+    Args:
+        tag: Optional version tag for decorated functions.
+
+    Returns:
+        A decorator that adds versioning capability to functions.
+    """
+    ...
+
+
+def version(
```
willbakst: Let's update this to be more similar to the updates made to the `@trace` decorator. It should return a `VersionedFunction` or `AsyncVersionedFunction` and follow a similar patter with `VersionDecorator` etc.?

sdks/python/src/lilypad/remote/__init__.py
koxudaxi: I forgot it. I just deleted it.

sdks/python/src/lilypad/tracing.py
koxudaxi: Oh, I made mistake. thank you.

sdks/python/src/lilypad/types.py
koxudaxi: I've moved the duplicated parts to the Base class as well.
The reason for separating Async and Sync here is to resolve the complexity of type inference.
While it would likely be possible to use a single class for both async and sync operations using something like Awaitable, I decided to provide separate classes for sync and async because it would become difficult for users to determine whether the result they obtained is synchronous or asynchronous.
What do you think about this?

sdks/python/src/lilypad/types.py
koxudaxi: I just moved it to version module

sdks/python/src/lilypad/versioning/decorator.py
koxudaxi: I applied the same structure to VersionedDecorator.

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,364 @@
...
+    Sequence,
+    Mapping,
+)
+from typing_extensions import TypeIs, TypeAlias
+
+P = ParamSpec("P")
+R = TypeVar("R", covariant=True)
+
+
+class JsonableObject(Protocol):
```
willbakst: Let's drop this, we don't need `JsonableObject`

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,364 @@
...
+)
+"""Simple type alias for JSON-serializable types."""
+
+AnnotationData = dict[str, Jsonable]
+
+
+def _fn_is_async(
+    fn: SyncFunction[P, R] | AsyncFunction[P, R],
+) -> TypeIs[AsyncFunction[P, R]]:
+    """[MISSING DOCSTRING]"""
```
willbakst: nit: let's fill in these missing docstrings. I was using them as a placeholder when pair programming

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,364 @@
...
+
+    async def __call__(self, *args: P.args, **kwargs: P.kwargs) -> R:
+        """[MISSING DOCSTRING]"""
+        ...
+
+
+@dataclass(kw_only=True)
+class BaseTracedFunction:
+    tags: tuple[str, ...]
```
willbakst: nit: let's give these fields docstrings? and newline separation?

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,364 @@
...
+    _is_async: bool
+    _qualified_name: str = field(init=False)
+    _module_name: str = field(init=False)
+
+
+@dataclass(kw_only=True)
+class TracedFunction(Generic[P, R], BaseTracedFunction):
+    """[MISSING DOCSTRING]"""
+
+    fn: SyncFunction[P, R]
```
willbakst: ditto here re: docstrings + newlines

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,364 @@
...
+            *args: Positional arguments to pass to the wrapped function.
+            **kwargs: Keyword arguments to pass to the wrapped function.
+
+        Returns:
+            The result from the wrapped function.
+        """
+        raise NotImplementedError("TracedFunction.__call__ not yet implemented")
+
+    def wrap(self, *args: P.args, **kwargs: P.kwargs) -> TraceResult[R]:
+        """Execute the wrapped function with tracing enabled.
```
willbakst: nit: the tracing will still be enabled for `__call__`, so this should say something like this instead:

Return a wrapper around the executed function's result for trace utilities.

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,364 @@
...
+            A TraceResult containing the function response and trace context.
+        """
+        raise NotImplementedError("TracedFunction.wrap not yet implemented")
+
+
+@dataclass(kw_only=True)
+class AsyncTracedFunction(Generic[P, R], BaseTracedFunction):
+    """[MISSING DOCSTRING]"""
+
+    fn: AsyncFunction[P, R]
```
willbakst: ditto here re: docstrings + newlines

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,364 @@
...
+            *args: Positional arguments to pass to the wrapped function.
+            **kwargs: Keyword arguments to pass to the wrapped function.
+
+        Returns:
+            The result from the wrapped async function.
+        """
+        raise NotImplementedError("AsyncTracedFunction.__call__ not yet implemented")
+
+    async def wrap(self, *args: P.args, **kwargs: P.kwargs) -> AsyncTraceResult[R]:
+        """Execute the wrapped async function with tracing enabled.
```
willbakst: ditto

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,364 @@
...
+
+@dataclass(kw_only=True)
+class TracedFunction(Generic[P, R], BaseTracedFunction):
+    """[MISSING DOCSTRING]"""
+
+    fn: SyncFunction[P, R]
+    _is_async: bool = field(default=False, init=False)
+
+    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> R:
+        """Call the wrapped function directly without tracing.
```
willbakst: this should trace the function. calling `.wrap` is just so you can get the wrapped trace result for trace utilities like `.annotate` or `.assign`

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,364 @@
...
+
+@dataclass(kw_only=True)
+class AsyncTracedFunction(Generic[P, R], BaseTracedFunction):
+    """[MISSING DOCSTRING]"""
+
+    fn: AsyncFunction[P, R]
+    _is_async: bool = field(default=True, init=False)
+
+    async def __call__(self, *args: P.args, **kwargs: P.kwargs) -> R:
+        """Call the wrapped async function directly without tracing.
```
willbakst: ditto

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,364 @@
...
+    async def assign(self, *emails: str) -> None:
+        """Assign the trace to specific users by email.
+
+        Args:
+            *emails: Variable number of email addresses to assign.
+        """
+        raise NotImplementedError("AsyncTraceResult.assign not yet implemented")
+
+
+class TraceResult(BaseTraceResult[R]):
```
willbakst: nit: put this before it's async counterpart since that's what we do elsewhere (putting sync first)

sdks/python/src/lilypad/types.py
willbakst: Sync and Async makes sense. It mostly seems like this entire file can be deleted since it's functionality already exists in the `tracing.py` file?

sdks/python/src/lilypad/tracing.py
willbakst, line 1: nit: let's rename this to `traces.py`?

sdks/python/src/lilypad/versioning/types.py
willbakst, line 1: let's just merge this with `decorator.py` into a single `versioning.py` file for now? We can always refactor as we implement and see a need for refactoring.

sdks/python/src/lilypad/versioning/decorator.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+@dataclass(kw_only=True)
+class AsyncVersionedFunction(Generic[P, R], BaseVersionedFunction):
+    """Wrapper for asynchronous functions with versioning capabilities."""
+
+    fn: AsyncFunction[P, R]
+    _is_async: bool = field(default=True, init=False)
+
+    def __post_init__(self) -> None:
```
willbakst: shouldn't this be on `BaseVersionedFunction` so we don't have to duplicate it? Also shouldn't this be the same for `TracedFunction`? Perhaps we can pull a base class for versioned functions from traced functions since a versioned function is a super class of a traced function?

sdks/python/src/lilypad/versioning/decorator.py
```diff
@@ -0,0 +1,327 @@
...
+            A versioned function wrapper.
+        """
+        if _fn_is_async(fn):
+            return AsyncVersionedFunction(fn=fn, tag=self.tag)
+        else:
+            return VersionedFunction(fn=fn, tag=self.tag)
+
+
+@dataclass(kw_only=True)
+class BaseVersionedFunction:
```
willbakst: yeah doubling down on my other comment, let's make `VersionedFunction` a super class of `TracedFunction`

sdks/python/src/lilypad/versioning/decorator.py
```diff
@@ -0,0 +1,327 @@
...
+
+class AsyncFunction(Protocol[P, R]):
+    """Protocol for asynchronous callable functions."""
+
+    async def __call__(self, *args: P.args, **kwargs: P.kwargs) -> R:
+        """Call the async function."""
+        ...
+
+
+def _fn_is_async(
```
willbakst: this is duplicated

sdks/python/src/lilypad/versioning/decorator.py
```diff
@@ -0,0 +1,327 @@
...
+from typing_extensions import TypeIs
+
+from ..tracing import TraceResult, AsyncTraceResult
+from .types import VersionRef
+
+P = ParamSpec("P")
+R = TypeVar("R", covariant=True)
+
+
+class SyncFunction(Protocol[P, R]):
```
willbakst: these are duplicated. we should DRY this code

sdks/python/src/lilypad/versioning/decorator.py
```diff
@@ -0,0 +1,327 @@
...
+            ref: Version reference (tag, hash, or semantic version).
+            *args: Positional arguments for the function.
+            **kwargs: Keyword arguments for the function.
+
+        Returns:
+            The function result from the specified version.
+        """
+        raise NotImplementedError("VersionedFunction.run_version not yet implemented")
+
+    def run_version_wrap(
```
willbakst: with the inheritance update, this should have the `wrap` method from `TracedFunction` as well, so we'll have `__call__`, `wrap`, `run_version`, and `run_version_wrap`.

I wonder if we can make this naming better. For example, I wonder if it should be `wrapped` instead of `wrap`?

I also wonder if it should be `run` instead of `run_version` since it reads cleaner with kwargs to have `my_fn.run(version="...")`. We could even make `version` optional where if `None` it simply runs `__call__`?

Same with `run_version_wrap` which would become `run_wrapped`. Same thing with `version` being optional where if `None` this simply runs `wrapped`?

In this case, I wonder if `VersionedFunction` should inherit from `BaseTracedFunction` instead of `TracedFunction` and then `TracedFunction` should implement `wrapped` without an arg and `VersionedFunction` should implement `wrapped` with the `version: VersionRef` arg?

We could then have `BaseTracedFunction` and `BaseAsyncTracedFunction` that are actually `ABC` and have `__call__` as an abstract method where the `TracedFunction`/`AsyncTracedFunction` and `VersionedFunction`/`AsyncVersionedFunction` variants inherit from their batching base class. Then the versioned function variants could additionally implement the `.run` method. I'm not opposed to `.run_version` to be more explicit in this case, although I think `__call__`  + `wrapped(version="...")` and `run(version="...")` reads more cleanly so long as we require kwargs with `*`.

sdks/python/src/lilypad/versioning/decorator.py
koxudaxi: The reason I didn't include `__post_init__` in `BaseVersionedFunction` is that `BaseVersionedFunction` doesn't define `fn`. There was no clean way to overload the Generic types.
However, since I'm already calling `getattr` to manipulate `fn`, I might as well call `fn` itself using `getattr` too.

sdks/python/src/lilypad/tracing.py
koxudaxi: I wrote all docstrings in up stack. I will copy it to this stack.

koxudaxi: Done

sdks/python/src/lilypad/versioning/decorator.py
koxudaxi: Done

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,526 @@
...
+    ParamSpec,
+    TypeVar,
+    overload,
+    Protocol,
+    Sequence,
+    Mapping,
+)
+from typing_extensions import TypeIs, TypeAlias
+
+from lilypad._internal.utils import get_qualified_name
```
willbakst, line 9: we're now making this util public when we shouldn't be, instead we should be importing `from lilypad._internal import utils as _utils` to keep it private

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,526 @@
...
+
+P = ParamSpec("P")
+R = TypeVar("R", covariant=True)
+
+
+Jsonable: TypeAlias = (
```
willbakst: nit: this should probably live in the `types.py` module

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,526 @@
...
+        """Get the function UUID if available.
+
+        Returns:
+            The function UUID or None if not available.
+        """
+        return self._function_uuid
+
+
+class TraceResult(BaseTraceResult[R]):
+    """Per-call handle returned by sync .wrapped() methods.
```
willbakst: nit: code should be in backticks. same elsewhere

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,526 @@
...
+    Args:
+        fn: The function to check.
+
+    Returns:
+        True if the function is asynchronous, False otherwise.
+    """
+    return inspect.iscoroutinefunction(fn)
+
+
+class BaseTraceResult(Generic[R]):
```
willbakst: nit: using `R` here is a bit confusing, it should probably be something more like `CovariantT`? I feel like `R` should be for return type generics of functions?

willbakst: Let's just make this a frozen dataclass rather than implementing private getters ourselves.

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,526 @@
...
+)
+"""Simple type alias for JSON-serializable types."""
+
+AnnotationData = dict[str, Jsonable]
+
+
+def _fn_is_async(
+    fn: SyncFunction[P, R] | AsyncFunction[P, R],
+) -> TypeIs[AsyncFunction[P, R]]:
+    """Check if a function is asynchronous.
```
willbakst: nit: also no need for this verbose docstring since the type hinting already explains all of this. same elsewhere

sdks/python/src/lilypad/tracing.py
```diff
@@ -0,0 +1,526 @@
...
+            **kwargs: Keyword arguments to pass to the wrapped function.
+
+        Returns:
+            An AsyncTraceResult containing the function response and trace context.
+        """
+        ...
+
+
+@dataclass(kw_only=True)
+class TracedFunction(BaseTracedFunction[P, R]):
```
willbakst: nit: let's move this right after it's base class to the sync/async classes live together

sdks/python/src/lilypad/versioning.py
```diff
@@ -0,0 +1,345 @@
...
+            A versioned function wrapper.
+        """
+        if _fn_is_async(fn):
+            return AsyncVersionedFunction(fn=fn, tag=self.tag, tags=())
+        else:
+            return VersionedFunction(fn=fn, tag=self.tag, tags=())
+
+
+@dataclass(kw_only=True)
+class BaseVersionedFunction:
```
willbakst: this is still duplicated heavily with traced functions and should be DRY

sdks/python/src/lilypad/versioning.py
```diff
@@ -0,0 +1,345 @@
...
+
+    def get_version(self) -> str:
+        """Get the current version identifier.
+
+        Returns:
+            The current version ID as a string.
+        """
+        raise NotImplementedError("VersionedFunction.get_version not yet implemented")
+
+    def run(self, *args: Any, **kwargs: Any) -> R:
```
willbakst: this is missing how the user specifies the version ID? I feel like calling `run` on a `VersionedFunction` with a specific version ID should then return a VersionedFunction that can be called with the given arguments? Otherwise this is the same as `__call__`.

willbakst (approved):

willbakst:
### Merge activity

* **Sep 10, 4:28 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/lilypad/692).

PR #702, opened 2025-09-19
feat: Implement Google OpenTelemetry

koxudaxi:
### TL;DR

Added Google GenAI instrumentation for telemetry and tracing in the Lilypad Python SDK.

### What changed?

- Implemented `GoogleGenAIInstrumentor` class to instrument Google GenAI clients for telemetry collection
- Added support for both synchronous and asynchronous API calls
- Implemented handling for streaming responses
- Added support for processing various content types including user, system, and assistant messages
- Added extraction of tool/function calls from responses
- Implemented standardized attribute and event extraction for OpenTelemetry spans
- Added comprehensive test suite with VCR cassettes for replay-based testing

### How to test?

1. Install the Google GenAI Python client:
   ```
   pip install google-generativeai
   ```

2. Create and instrument a client:
   ```python
   from google.genai import Client
   from lilypad._internal.otel.google_genai import instrument_google_genai
   
   client = Client(api_key="your-api-key")
   instrument_google_genai(client)
   
   # Use the client as normal
   response = client.models.generate_content(
       model="gemini-1.5-flash",
       contents="Hello, world!",
   )
   ```

3. Run the test suite:
   ```
   pytest sdks/python/tests/lilypad/_internal/otel/test_google_genai.py -v
   ```

### Why make this change?

This change adds support for instrumenting Google GenAI clients, allowing for consistent telemetry collection across different LLM providers. This enables users to track and analyze their Google GenAI API usage with the same tooling and metrics as other supported providers, providing a unified observability experience within the Lilypad ecosystem.


koxudaxi:
* **#703** <a href="https://app.graphite.com/github/pr/Mirascope/lilypad/703?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#702** <a href="https://app.graphite.com/github/pr/Mirascope/lilypad/702?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Mirascope/lilypad/702?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#717** <a href="https://app.graphite.com/github/pr/Mirascope/lilypad/717?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#718** <a href="https://app.graphite.com/github/pr/Mirascope/lilypad/718?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#713** <a href="https://app.graphite.com/github/pr/Mirascope/lilypad/713?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#708** <a href="https://app.graphite.com/github/pr/Mirascope/lilypad/708?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#707** <a href="https://app.graphite.com/github/pr/Mirascope/lilypad/707?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#706** <a href="https://app.graphite.com/github/pr/Mirascope/lilypad/706?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v1`


<h2></h2>

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+    GenAIResponseAttributes,
+    Message,
+    MessageEvent,
+    SystemMessageEvent,
+    ToolCall,
+    ToolCallDelta,
+    UserMessageEvent,
+)
+
+_GOOGLE_GENAI_SYSTEM = "google_genai"
```
willbakst: nit: is there no `gen_ai_attributes.GenAiSystemValues.OPENAI.value` equivalent?

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+
+GoogleGenAIClient: TypeAlias = Client
```
willbakst: nit: I think just `GoogleClient` would be fine?

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+    # REQUIRED
+    model: Required[str]
+    contents: Required[ContentUnion | Iterable[ContentUnion]]
+
+    # NOT REQUIRED
+    config: GenerateContentConfig | GenerateContentConfigDict | None
+
+
+def _maybe_file(value: PartInput) -> File | None:
+    """Returns a File when the input represents file metadata."""
```
willbakst: nit: update to `File` in backticks

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+
+        return response_attributes, choice_deltas
+
+
+def instrument_google_genai(
+    client: GoogleGenAIClient,
+) -> None:
+    """Instruments a Google GenAI client for telemetry collection.
+
+    Args:
```
willbakst: nit: this `Args` section is unnecessary since the type hint provides the same value

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+
+    @staticmethod
+    def _process_chunk(
+        chunk: GenerateContentResponse,
+    ) -> tuple[GenAIResponseAttributes, list[ChoiceDelta]]:
+        """Returns response attributes and choice deltas from Google GenAI streaming chunk."""
+        response_attributes = GenAIResponseAttributes()
+        choice_deltas: list[ChoiceDelta] = []
+
+        response_attributes.GEN_AI_RESPONSE_ID = None
```
willbakst: I don't remember why I did this for OpenAI and Anthropic, but I feel like we should put these initial values directly in the constructor and only manually set attributes in the conditional cases?

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+
+        if chunk.usage_metadata:
+            response_attributes.GEN_AI_USAGE_INPUT_TOKENS = (
+                chunk.usage_metadata.prompt_token_count
+            )
+            response_attributes.GEN_AI_USAGE_OUTPUT_TOKENS = (
+                chunk.usage_metadata.candidates_token_count
+            )
+
+        if chunk.candidates:
```
willbakst: let's remove this extra level of indentation with a `or []` usage in the iterator below

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+            finish_reasons = []
+            for candidate in chunk.candidates:
+                index = candidate.index or 0
+                finish_reason = None
+
+                if candidate.finish_reason:
```
willbakst: Is there a chance that `finish_reason` below will be `None` if `candidate.finish_reason` evaluates to `True`?

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+
+                content = candidate.content
+                text_delta = None
+                tool_call_deltas = None
+
+                if content and content.parts:
+                    for part in content.parts:
+                        text, func_call = _part_text_and_function_call(part)
+                        if text:
+                            text_delta = text
```
willbakst: seems weird here to be assigning for each iteration through parts? is there an assumption here that there will only ever be a single part with a text delta ?

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+                    ChoiceDelta(
+                        system=_GOOGLE_GENAI_SYSTEM,
+                        index=index,
+                        content=text_delta,
+                        tool_calls=tool_call_deltas,
+                        finish_reason=finish_reason,
+                    )
+                )
+
+            if finish_reasons:
```
willbakst: I'm not seeing this conditional for OpenAI, so we should match these. Seems like the conditional is likely the path that makes the most sense to me, but I need to refresh on the types.

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+        return message_events
+
+    @staticmethod
+    def _process_response(
+        response: GenerateContentResponse
+        | Iterator[GenerateContentResponse]
+        | AsyncIterator[GenerateContentResponse],
+    ) -> tuple[list[ChoiceEvent], GenAIResponseAttributes]:
+        """Returns the choice events list and response attributes from Google GenAI response."""
+        if isinstance(response, Iterator | AsyncIterator):
```
willbakst: What's up with these conditionals? Shouldn't this method only ever receive `GenerateContentResponse`?

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+            return [], GenAIResponseAttributes()
+
+        if not isinstance(response, GenerateContentResponse):
+            return [], GenAIResponseAttributes()
+
+        choice_events: list[ChoiceEvent] = []
+        finish_reasons: list[str] = []
+
+        if response.candidates:
```
willbakst: nit: same thing here with `or []`

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+
+                if candidate.finish_reason:
+                    finish_reason = candidate.finish_reason.value
+
+                content = candidate.content
+                text = None
+                tool_calls = None
+
+                if content:
+                    text, tool_calls = _extract_content_from_parts(content.parts)
```
willbakst: seems like there's possible overlap here with the separate method used when process chunks `_part_text_and_function_call`

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+                if finish_reason:
+                    finish_reasons.append(finish_reason)
+
+        response_attributes = GenAIResponseAttributes(
+            GEN_AI_RESPONSE_ID=None,
+            GEN_AI_RESPONSE_MODEL=response.model_version,
+            GEN_AI_RESPONSE_FINISH_REASONS=finish_reasons or None,
+        )
+
+        if response.usage_metadata:
```
willbakst: we should match how we do this for other providers structurally

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+
+    @staticmethod
+    def _process_messages(
+        kwargs: _GoogleGenAIKwargs,
+    ) -> list[MessageEvent]:
+        """Returns standardized message events converted from Google GenAI contents."""
+        message_events: list[MessageEvent] = []
+        contents = kwargs.get("contents")
+
+        normalized_contents = _GoogleGenAIInstrumentor._normalize_message_contents(
```
willbakst: why not just make this another private utility method rather than a static or class method?

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+            contents
+        )
+
+        for content in normalized_contents:
+            role, text, tool_calls = _process_content(content)
+
+            if role == "system":
```
willbakst: we should use a match block here to match other provider implementations

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+        """Returns standardized message events converted from Google GenAI contents."""
+        message_events: list[MessageEvent] = []
+        contents = kwargs.get("contents")
+
+        normalized_contents = _GoogleGenAIInstrumentor._normalize_message_contents(
+            contents
+        )
+
+        for content in normalized_contents:
+            role, text, tool_calls = _process_content(content)
```
willbakst: seems like a waste to normalize the content rather than just have `_process_content` handle that

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+
+@overload
+def _ensure_part(
+    value: PartInput,
+    *,
+    keep_primitives: Literal[True],
+) -> Part | str | PIL.Image.Image | File: ...
+
+
+def _ensure_part(
```
willbakst: This seems like a bunch of extra compute creating (and validating) parts when not necessary. We should think through this.

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,455 @@
...
+                        "finish_reason": "STOP",
+                    },
+                },
+            ],
+        }
+    )
+
+
+@pytest.mark.vcr()
+def test_google_genai_with_system(
```
willbakst: nit: let's just put a system message in the simple case for merged tests

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,455 @@
...
+                        "message": '{"role":"assistant","content":"The available tools lack the ability to access real-time weather information.  Therefore, I cannot answer your question.\\n"}',
+                        "finish_reason": "STOP",
+                    },
+                },
+            ],
+        }
+    )
+
+
+def test_google_genai_already_instrumented(client: Client) -> None:
```
willbakst: match this test name with the openai and anthropic ones i.e. "idempotent"

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,455 @@
...
+                        "content": "Count from 1 to 5",
+                    },
+                }
+            ],
+        }
+    )
+
+
+@pytest.mark.vcr()
+def test_google_genai_with_tool_calls(
```
willbakst: let's do tool reinsertion testing rather than just a tool call test (since the tool call definitions aren't traced it seems)

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,455 @@
...
+                    },
+                },
+            ],
+        }
+    )
+
+
+@pytest.mark.vcr()
+@pytest.mark.asyncio
+async def test_google_genai_async(
```
willbakst: let's pair the sync/async versions of each test rather than having all sync and then all async

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,573 @@
...
+):
+    """Google GenAI client instrumentor for telemetry and tracing."""
+
+    @staticmethod
+    def _get_request_attributes(
+        kwargs: _GoogleGenAIKwargs,
+        client: GoogleGenAIClient,
+    ) -> GenAIRequestAttributes:
+        """Returns request attributes extracted from Google GenAI kwargs."""
+        attributes = GenAIRequestAttributes(
```
willbakst: Hmm, rather than constructing the object and then setting values, we should do what we do for OpenAI and get all the values that need special handling and then return a single object construction. This makes me wonder if we should make these dataclasses frozen

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,419 @@
...
+        elif isinstance(part_like, Part):
+            if part_like.file_data and part_like.file_data.mime_type:
+                text = f"[file:{part_like.file_data.mime_type}]"
+            elif part_like.inline_data and part_like.inline_data.mime_type:
+                text = f"[inline:{part_like.inline_data.mime_type}]"
+
+    return text, tool_call, tool_messages
+
+
+def _summarize_parts(
```
willbakst: nit: rename to `_combine_parts`

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,419 @@
...
+        role = _normalize_role(content.role, bool(tool_calls))
+        return role, text, tool_calls or None, tool_messages
+
+    elif isinstance(content, dict) and "parts" in content:
+        text, tool_calls, tool_messages = _summarize_parts(content.get("parts"))
+        role = _normalize_role(content.get("role"), bool(tool_calls))
+        return role, text, tool_calls or None, tool_messages
+
+    else:
+        assert not isinstance(content, Iterable) or isinstance(
```
willbakst: We shouldn't have assertions like this

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,419 @@
...
+        kwargs: _GoogleGenAIKwargs,
+        client: GoogleClient,
+    ) -> GenAIRequestAttributes:
+        """Return span attributes derived from the request parameters."""
+        raw_config = kwargs.get("config")
+        config: GenerateContentConfig | None
+        if raw_config is None or isinstance(raw_config, GenerateContentConfig):
+            config = raw_config
+        else:
+            config = GenerateContentConfig.model_validate(raw_config)
```
willbakst: We should flip this to use `GenerateContentConfigDict` as the unified type so we don't validate a valid config? This will also clean up the below since `.get` on a dictionary returns `None` by default if the key isn't present.

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,419 @@
...
+            if role == "assistant":
+                message_events.append(
+                    AssistantMessageEvent(
+                        system=_GOOGLE_GENAI_SYSTEM,
+                        content=text,
+                        tool_calls=tool_calls,
+                    )
+                )
+            else:
+                if text is not None or tool_calls or tool_messages:
```
willbakst: What is this conditional? It seems like if `tool_calls` evaluates to `True` that it should be an assistant message? But that's handled above already?

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,419 @@
...
+            message_events.extend(tool_messages)
+
+        return message_events
+
+    @staticmethod
+    def _process_response(
+        response: GenerateContentResponse,
+    ) -> tuple[list[ChoiceEvent], GenAIResponseAttributes]:
+        """Summarize a non-streaming response into events and attributes."""
+        if not isinstance(response, GenerateContentResponse):
```
willbakst: This goes against the type hint? When would it not be a `GenerateContentResponse`? In such a case, what would it be? That should be included in the type if possible.

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,419 @@
...
+def _summarize_parts(
+    parts: Iterable[PartLike] | None,
+) -> tuple[str | None, list[ToolCall], list[ToolMessageEvent]]:
+    """Aggregate text and tool metadata across the supplied part collection."""
+    text_segments: list[str] = []
+    tool_calls: list[ToolCall] = []
+    tool_messages: list[ToolMessageEvent] = []
+
+    for part in parts or []:
+        text, tool_call, messages = _parse_part(part)
```
willbakst: nit: should be `tool_messages` not `messages`

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,419 @@
...
+        elif "executable_code" in part_like or "executableCode" in part_like:
+            code_data = part_like.get("executable_code") or part_like.get(
+                "executableCode"
+            )
+            language = (
+                code_data.get("language") if isinstance(code_data, dict) else "unknown"
+            )
+            text = f"[code:{language}]"
+        elif part_like.get("thought"):
+            text = "[thought]"
```
willbakst: Why are we just ignoring the entire thought?

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,419 @@
...
+        elif "video_metadata" in part_like or "videoMetadata" in part_like:
+            text = "[video]"
```
willbakst: This seems like we don't actually handle it? Should we just `raise NotImplementedError` for now until we implement it?

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,419 @@
...
+                else GoogleFunctionResponse.model_validate(function_response)
+            )
+            tool_messages.extend(_tool_messages_from_response(response_obj))
+    elif isinstance(part_like, File):
+        text = f"[file:{part_like.mime_type or 'unknown'}]"
+    elif isinstance(part_like, PIL.Image.Image):
+        image_format = getattr(part_like, "format", None)
+        text = f"[image/{image_format.lower()}]" if image_format else "[image]"
+
+    if text is None:
```
willbakst: Why is this wrapped in an additional `text is None` block? Why not just have it as part of the above block?

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,968 @@
...
+                        "finish_reason": "STOP",
+                    },
+                },
+            ],
+        }
+    )
+
+
+@pytest.mark.vcr()
+def test_google_genai_tool_message_resinsertion(
```
willbakst: This is not testing "reinsertion" it's just testing a call with tool calls. We should be testing the next step after that which is handling when the user provides the outputs of the tools (i.e. reinserts them)?

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,968 @@
...
+                        "finish_reason": "STOP",
+                    },
+                },
+            ],
+        }
+    )
+
+
+@pytest.mark.vcr()
+def test_google_genai_tool_call(
```
willbakst: Aha, this is the reinsertion test, and the other is the tool call. I'm not sure we need both. Can we merge. these and still retain coverage?

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,968 @@
...
+                        "tool_calls": '[{"id":"","type":"function","function":{"name":"get_weather","arguments":"{\\"location\\":\\"New York\\"}"}}]',
+                    },
+                },
+            ],
+        }
+    )
+
+
+@pytest.mark.vcr()
+def test_google_genai_multi_turn(
```
willbakst: Does this provide additional coverage that other tests don't cover? If anything I would just merge this with the simple test?

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,968 @@
...
+                    },
+                },
+            ],
+        }
+    )
+
+
+@pytest.mark.vcr()
+@pytest.mark.asyncio
+async def test_google_genai_async_streaming(
```
willbakst: We should pair this structurally with it's sync sibling

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,968 @@
...
+                        "content": "Count from 1 to 5",
+                    },
+                }
+            ],
+        }
+    )
+
+
+@pytest.mark.vcr()
+def test_google_genai_dict_inline_data(
```
willbakst: Why not just merge this into a single test for both possible types of inline data (using both in the same test)?

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,968 @@
...
+                        "finish_reason": "STOP",
+                    },
+                },
+            ],
+        }
+    )
+
+
+@pytest.mark.vcr()
+def test_google_genai_iterable_parts(
```
willbakst: I feel like we should have a single test that just has all the different various parts and snapshot asserts we handle them correctly rather than having a single test per?

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,968 @@
...
+                    "attributes": {
+                        "gen_ai.system": "gcp.gen_ai",
+                        "content": "[inline:video/gif]",
+                    },
+                },
+                {
+                    "name": "gen_ai.user.message",
+                    "attributes": {
+                        "gen_ai.system": "gcp.gen_ai",
+                        "content": "What do you see in this animation?",
```
willbakst: I'm not actually seeing the video stuff here?

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,968 @@
...
+                    },
+                },
+                {
+                    "name": "gen_ai.choice",
+                    "attributes": {
+                        "gen_ai.system": "gcp.gen_ai",
+                        "index": 0,
+                        "message": '{"role":"assistant","content":"Please provide me with the animation! I need to see it in order to tell you what I see. I can analyze animations based on their visuals, motion, and any elements present.\\n"}',
```
willbakst: Apparently it's not seeing it either?

sdks/python/tests/lilypad/_internal/otel/test_google.py
willbakst, line 1: High level this is much improved. I still think we can merge a bunch of the tests, specifically around part handling?

sdks/python/src/lilypad/_internal/otel/google.py
koxudaxi: This check is necessary because generate_content_stream returns an async generator that doesn't satisfy the AsyncStream  protocol (has aclose not close), so it reaches _process_response() and would fail accessing .candidates.

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,477 @@
...
+            )
+            language = (
+                language_raw.value
+                if language_raw and hasattr(language_raw, "value")
+                else str(language_raw)
+                if language_raw
+                else "unknown"
+            )
+            text = f"[code:{language}]"
+        elif part_like.get("thought"):
```
willbakst: Is the `thought` field not the text of the thought?

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,477 @@
...
+
+        function_call = part_like.get("function_call") or part_like.get("functionCall")
+        if function_call:
+            google_call = (
+                function_call
+                if isinstance(function_call, GoogleFunctionCall)
+                else GoogleFunctionCall.model_validate(function_call)
+            )
+            tool_call = _tool_call_from_google(google_call)
+        function_response = part_like.get("function_response") or part_like.get(
```
willbakst, line 219: nit: put a newline above this

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,477 @@
...
+                else None
+            )
+            text = f"[inline:{mime or 'unknown'}]"
+
+        function_call = part_like.get("function_call") or part_like.get("functionCall")
+        if function_call:
+            google_call = (
+                function_call
+                if isinstance(function_call, GoogleFunctionCall)
+                else GoogleFunctionCall.model_validate(function_call)
```
willbakst: Why validate here? If it's not already an instance of `GoogleFunctionCall`, is it a dictionary? Seems like we should just have `_tool_call_from_google` handle that rather than constructing and validating the model like this?

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,477 @@
...
+        )
+        if function_response:
+            response_obj = (
+                function_response
+                if isinstance(function_response, GoogleFunctionResponse)
+                else GoogleFunctionResponse.model_validate(function_response)
+            )
+            tool_messages.extend(_tool_messages_from_response(response_obj))
+
+        thought_sig = part_like.get("thought_signature") or part_like.get(
```
willbakst: This should go with `thought` right?

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,477 @@
...
+                metadata_parts.append(f"fps={part_like.video_metadata.fps}")
+            text = (
+                f"[video:{','.join(metadata_parts) if metadata_parts else 'metadata'}]"
+            )
+
+        if part_like.function_call:
+            tool_call = _tool_call_from_google(part_like.function_call)
+        tool_messages.extend(_tool_messages_from_response(part_like.function_response))
+
+        if part_like.thought_signature and not text and not tool_call:
```
willbakst: this should go together with `thought` right? so they are together

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,477 @@
...
+):
+    """Google GenAI client instrumentor for telemetry and tracing."""
+
+    @staticmethod
+    def _get_request_attributes(
+        kwargs: _GoogleGenAIKwargs,
+        client: GoogleClient,
+    ) -> GenAIRequestAttributes:
+        """Return span attributes derived from the request parameters."""
+        raw_config = kwargs.get("config")
```
willbakst: This can simpler:

```
config: GenerateContentConfig | GenerateContentConfigDict = kwargs.get("config") or {}
if isinstance(config, GenerateContentConfig):
    config = cast(GenerateContentConfigDict, config.model_dump())
```

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,477 @@
...
+            config = cast(GenerateContentConfigDict, raw_config.model_dump())
+        else:
+            config = raw_config
+
+        return GenAIRequestAttributes(
+            GEN_AI_SYSTEM=_GOOGLE_GENAI_SYSTEM,
+            SERVER_ADDRESS="generativelanguage.googleapis.com",
+            SERVER_PORT=443,
+            GEN_AI_REQUEST_MODEL=kwargs.get("model"),
+            GEN_AI_REQUEST_TEMPERATURE=config.get("temperature") if config else None,
```
willbakst: If we remove the possibility of `None` as an option for `config` as in my comment above, we can drop all of these conditions and ternaries.

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,477 @@
...
+        role = _normalize_role(content.role, bool(tool_calls))
+        return role, text, tool_calls or None, tool_messages
+
+    elif isinstance(content, dict) and "parts" in content:
+        text, tool_calls, tool_messages = _combine_parts(content.get("parts"))
+        role = _normalize_role(content.get("role"), bool(tool_calls))
+        return role, text, tool_calls or None, tool_messages
+
+    else:
+        text, tool_call, tool_messages = _parse_part(cast(PartLike, content))
```
willbakst: Why do we need to cast here?

sdks/python/src/lilypad/_internal/otel/google.py
willbakst: This seems like a bug that we should handle so that `_process_response` only ever gets a `GenerateContentResponse` object?

sdks/python/tests/lilypad/_internal/otel/test_google.py
willbakst: I still think we can simplify. Let's go through this together.

sdks/python/src/lilypad/_internal/otel/google.py
koxudaxi: `thought` is just the boolean flag indicating the part represents model thinking. The actual thought text stays in the text field.

sdks/python/src/lilypad/_internal/otel/google.py
koxudaxi: What do you mean? I have changed related codes.

sdks/python/src/lilypad/_internal/otel/google.py
koxudaxi: I have updated it.

sdks/python/src/lilypad/_internal/otel/google.py
koxudaxi: We do some conversion before, so at runtime the type should be correct. But Pyright thinks Iterator[Parts] and others are still there, so we get errors. I will try to find a good solution.

koxudaxi: OK, I fixed the condition without cast.

sdks/python/src/lilypad/_internal/otel/google.py
koxudaxi: make sense. I fixed the issue for condition `aclose` method for `GenerateContentResponse `

sdks/python/tests/lilypad/_internal/otel/test_google.py
koxudaxi: I updated the tests

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,600 @@
...
+        contents="List two benefits of observability.",
+        config=GenerateContentConfig(max_output_tokens=40),
+    )
+    collected = []
+    async for chunk in stream:
+        collected.append(chunk)
+    assert isinstance(stream, AsyncStreamWrapper)
+    assert isinstance(stream.stream, AsyncIteratorStream)
+    assert aiter(stream.stream) is stream.stream
+    await stream.close()
```
koxudaxi, line 1072: The above four lines are coverage additions corresponding to the fix for the comment below. I couldn't think of a very reasonable test to achieve 100% coverage. However, I couldn't think of any other good implementation either, so please let me know if there's a better approach.

https://app.graphite.dev/github/pr/Mirascope/lilypad/702/feat-Implement-Google-OpenTelemetry#comment-PRRC_kwDOMFfGHM6PE2Qr

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,468 @@
...
+            GEN_AI_RESPONSE_MODEL=chunk.model_version,
+            GEN_AI_RESPONSE_FINISH_REASONS=finish_reasons or None,
+            GEN_AI_USAGE_INPUT_TOKENS=usage.prompt_token_count if usage else None,
+            GEN_AI_USAGE_OUTPUT_TOKENS=usage.candidates_token_count if usage else None,
+        )
+
+        return response_attributes, choice_deltas
+
+
+def instrument_google_genai(
```
willbakst: nit: we should rename this to `instrument_google`

sdks/python/src/lilypad/_internal/otel/base_instrumentor.py
```diff
@@ -197,6 +198,13 @@ async def wrapper(
                             span, response, self._process_chunk
                         )
                         return cast(ResponseT, wrapped_async_stream)
+                    if isinstance(response, AsyncIterator):
```
willbakst: Seems like we could just merge with the `AsyncStream` case given how similar they are. It appears like the only difference is the `close` vs. `aclose` issue.

sdks/python/src/lilypad/_internal/otel/stream_wrappers.py
```diff
@@ -299,7 +318,17 @@ async def __aexit__(
 
     async def close(self) -> None:  # pragma: no cover
         """Close the async stream and perform cleanup."""
-        await self.stream.close()
+        close_candidate = getattr(self.stream, "close", None)
```
willbakst: It feels like we're covering both bases here rather than picking a path? In one case we update so it will always have a `close` method, but then here we detect whether it's `close` or `aclose`. It seems like we should just find a way to type the async stream case such that `AsyncStreamWrapper` can just handle both cases like this is implemented.

sdks/python/tests/lilypad/_internal/otel/test_google.py
willbakst: I think simply merging the paths as mentioned in my other comments should cover this. I will see if that's the case.

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,600 @@
...
+                        "message": '{"role":"assistant","content":"Here are two key benefits of observability:\\n\\n1.  **Faster Problem Resolution:** Observability provides deep insights into the internal states of a system based on its outputs (metrics, logs, traces)."}',
+                        "finish_reason": "MAX_TOKENS",
+                    },
+                },
+            ],
+        }
+    )
+
+
+def test_google_genai_instrumentation_idempotent() -> None:
```
willbakst, line 1377: nit: let's make sure this matches all of the other providers exactly

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,600 @@
...
+                code_execution=ToolCodeExecution(),
+            )
+        ],
+        thinking_config=ThinkingConfig(include_thoughts=True),
+    )
+    return uploaded_file, image, config
+
+
+@pytest.mark.vcr()
+def test_google_genai_pydantic_content_and_parts(
```
willbakst: I fear this is a little too combined. The size of the snapshots makes it difficult to grok.

In this case, I would likely keep a simple sync/async case, a case with tool reinsertion, and a content part pydantic vs. dict set of tests.

sdks/python/tests/lilypad/_internal/test_closure.py
```diff
@@ -477,7 +477,7 @@ def aliased_module_import_fn(query: str) -> str:
 """
     )
     assert closure.dependencies == snapshot(
-        {"openai": {"version": "1.98.0", "extras": None}}
+        {"openai": {"version": "1.98.0", "extras": ["realtime"]}}
```
willbakst: Why are we now including the `"realtime"` extra for OpenAI when we weren't before? I think you might have it installed on your computer but it's not actually part of this pyproject virtual environment?

sdks/python/src/lilypad/_internal/otel/google.py
```diff
@@ -0,0 +1,468 @@
...
+    @staticmethod
+    def _process_chunk(
+        chunk: GenerateContentResponse,
+    ) -> tuple[GenAIResponseAttributes, list[ChoiceDelta]]:
+        """Summarize a streaming chunk into response attributes and choice deltas."""
+        choice_deltas: list[ChoiceDelta] = []
+        finish_reasons: list[str] = []
+
+        for candidate in chunk.candidates or []:
+            text, tool_calls, _ = _combine_parts(
```
willbakst, line 438: Just now noticing we don't seem to do anything around testing streaming tools.

We should add streaming tools tests across the providers

sdks/python/tests/lilypad/_internal/test_closure.py
koxudaxi: My local version also doesn't include "realtime". It seems Google detected the dependency and included realtime. I'll fix it.

koxudaxi: Because `websockets` is bundled by other extras (for example the google), the old logic thought `websockets` came from realtime even when realtime wasn’t requested.
With this change, extras that are already satisfied by other imported distributions are excluded from the counted extras.

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,532 @@
...
+    but VCR has better support for httpx.
+    """
+
+    # Force httpx by providing a custom transport
+    # This prevents Google SDK from using aiohttp for streaming
+    http_options = HttpOptions(
+        async_client_args={"transport": httpx.AsyncHTTPTransport()}
+    )
+
+    client = Client(api_key=google_genai_api_key, http_options=http_options)
```
koxudaxi, line 71: @teamdandelion @willbakst 
I found out why vcr.py + async + streaming + Google SDK doesn't work in Lilypad but works in Mirascope. It's because Lilypad's dependencies include aiohttp, while Mirascope's don't. The Google SDK uses aiohttp when it's available, and falls back to httpx when it's not. I was able to reproduce the issue where it hangs and doesn't consume the stream properly by adding aiohttp to the mirascope v2 branch and trying to record a cassette.
As a workaround, I changed it in conftest.py to use httpx instead of aiohttp.
This is likely the same issue:
https://github.com/kevin1024/vcrpy/issues/927

sdks/python/tests/lilypad/_internal/otel/test_google.py
koxudaxi: I merged the paths but coverage didn't improve. The issue is that `AsyncStreamWrapper.__aiter__()` returns self, so `AsyncIteratorStream.__aiter__()` never gets called.
This means we still need that artificial assertion assert `aiter(stream.stream)` is stream.stream for coverage, which defeats the purpose of merging.
I tried a few different approaches, but meeting the protocol requirements just forces us to add pointless test code.

sdks/python/src/lilypad/_internal/closure.py
willbakst, line 1: What is this change and why is it part of this PR?

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,1049 @@
...
+    assert isinstance(response, GenerateContentResponse)
+
+    spans = span_exporter.get_finished_spans()
+    assert len(spans) == 1
+
+    span_data = extract_span_data(spans[0])
+    attrs = span_data["attributes"]
+
+    assert attrs.pop("gen_ai.input.messages") == snapshot(
+        '[{"role":"user","parts":[{"type":"text","content":"Explain observability in one sentence."}]}]'
```
willbakst: Same comment here re: loading these for improved readability

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,1049 @@
...
+            "events": [],
+        }
+    )
+
+
+@pytest.mark.vcr()
+def test_google_genai_pydantic_parts(
+    client: Client,
+    span_exporter: InMemorySpanExporter,
+    content_test_artifacts,
```
willbakst: nit: missing type

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,1049 @@
...
+) -> None:
+    """Test Pydantic Content and Part with various types (text, inline_data, thought, file, image)."""
+    uploaded_file, image, config = content_test_artifacts
```
willbakst: 1. I don't see you using `config` here?
2. Let's maybe have these as separate fixtures with individual types? I find this better than have a joint fixture you unpack.

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,1049 @@
...
+
+
+@pytest.mark.vcr()
+def test_google_genai_dict_parts(
+    client: Client,
+    span_exporter: InMemorySpanExporter,
+    content_test_artifacts,
+) -> None:
+    """Test dict Content and Part with various types."""
+    uploaded_file, image, config = content_test_artifacts
```
willbakst: Same comments here as above and elsewhere

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,1049 @@
...
+        }
+    )
+
+
+@pytest.mark.vcr()
+def test_google_genai_dict_advanced_parts(
+    client: Client,
+    span_exporter: InMemorySpanExporter,
+) -> None:
+    """Test dict Part with video_metadata, executable_code, code_execution_result."""
```
willbakst, line 611: How come we have some artifacts above that are shared between the pydantic vs. dict tests but here we inline? Seems like we should have fixtures for the underlying data for each part so that we can share for all of these tests?

sdks/python/tests/lilypad/_internal/otel/test_google.py
```diff
@@ -0,0 +1,1049 @@
...
+            },
+            "status": {"status_code": "UNSET", "description": None},
+            "events": [],
+        }
+    )
+
+
+@pytest.mark.vcr()
+@pytest.mark.asyncio
+async def test_async_google_genai_streaming(
```
willbakst: As before, we should generally have the sync vs. async tests paired up and we should try to have similar / matching naming conventions between them. This makes it easier to ensure we have both sync and async tests for something and easy to assume when a test doesn't have an async variant (cause it won't be right next to its sync variant).

sdks/python/tests/lilypad/_internal/test_closure.py
willbakst: Ok this explains the change to closure. We should probably have that as a separate change below this stack item so it's isolated.

sdks/python/tests/lilypad/_internal/otel/test_anthropic.py
```diff
@@ -394,6 +394,57 @@ def test_anthropic_streaming_with_tools(
     )
 
 
+@pytest.mark.vcr()
+@pytest.mark.asyncio
+async def test_async_anthropic_streaming_with_tools(
```
willbakst, line 558: We should add this test alongside the missing OpenAI streaming with tools tests in a separate PR from this Google one.

sdks/python/src/lilypad/_internal/otel/google.py
willbakst: I also commented on the Anthropic test added. We should do that stuff separately from this PR down stack so it's easier to review / isolate.

sdks/python/tests/lilypad/_internal/otel/test_google.py
willbakst: Ok, I think this is fine given the edge case.

sdks/python/tests/lilypad/_internal/test_closure.py
koxudaxi: OK, I just create a new parent stack. https://app.graphite.dev/github/pr/Mirascope/lilypad/717/Add-Google-dependency

sdks/python/src/lilypad/_internal/closure.py
koxudaxi: I created a new stack for the changes.

PR #704, opened 2025-09-23
test: remove '_generated' from coverage

willbakst:

PR #705, opened 2025-09-24
test: Require 100% coverage for SDKs and achieve it

koxudaxi:
# Improve Python SDK test coverage to 100%

## What changed?

This PR enhances the Python SDK test coverage to reach 100% by:

1. Configuring pytest to enforce 100% code coverage in `pyproject.toml`
2. Adding `pragma: no cover` markers to code paths that are unreachable or not practical to test
3. Adding comprehensive tests for previously untested components:
   - Client initialization and caching
   - Settings management
   - Error handling
   - OTLP exporters and attribute conversion
   - Span context and serialization
   - Tracing decorators and versioning

## Why this change?

Complete test coverage ensures:
- Higher code quality and reliability
- Better documentation of edge cases
- Easier maintenance as all code paths are verified
- Protection against regressions when making changes

The added tests also serve as examples of how to use various SDK components, which is valuable for both users and maintainers.

## Testing

The PR includes extensive new test cases that verify all code paths, including edge cases and error handling scenarios. The pytest configuration now enforces 100% coverage, ensuring this standard is maintained going forward.

GitHub: bacalhau-project/lilypad


koxudaxi:
* **#705** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/705?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/705?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#704** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/704?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v1`


<h2></h2>

sdks/python/src/lilypad/_internal/closure.py
willbakst, line 1: I would argue that the closure code should have basely any instances of `# pragma: no cover`. I say this because it's a lot of code, so we should remove any and all unnecessary code where possible. If all of our tests pass, but they never touch all of these lines, I would assume either that our tests are falling short or our implementation is bloated.

I'm going to take the time to comb through each ignore to figure out where the issue lies.

sdks/python/tests/lilypad/_internal/otel/exporters/test_exporters.py
```diff
@@ -645,221 +659,616 @@ def test_full_pipeline_with_vcr(
...
+                trace_id=0x03030303030303030303030303030303,
+                span_id=0x0303030303030303,
+                is_valid=True,
+            )
+        ],
+        attributes=cast(SpanAttributesDict, {"object_value": UnsupportedAttribute()}),
+        resource_attrs={},
+    )
+
+    result = exporter.export([cast(ReadableSpan, span)])
```
willbakst: Ok, I think there's been a misunderstanding around the usage of mocks. Mocks are extremely useful for dummy data, and all of these dummy classes defeat the point. We might as well use mocks.

What we shouldn't do is use mocking to test our internal implementation.

To be more clear, there's one case where we use mocks and assert that our mock is the argument to some specific function call internally. This would be testing our implementaiton, which makes the tests brittle.

Instead, we should use mocks to assert that what we expect to happen happened (from the user's perspective, not the perspective of the author).

For example, if we were to create a `Mock(spec=ReadableSpan)`, we should be able to assert that said mock was in fact exported.

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,234 @@
...
+    get_sync_client,
+)
+from lilypad._internal.settings import (
+    CURRENT_SETTINGS,
+    settings as settings_override,
+)
+
+
+@pytest.fixture(autouse=True)
+def reset_client_caches() -> Generator[None, None, None]:
```
willbakst, line 22: nit: these methods are missing docstrings

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,234 @@
...
+    close_cached_clients()
+    yield
+    close_cached_clients()
+
+
+def _close_sync_client(client: Lilypad) -> None:
+    http_client = client._client_wrapper.httpx_client.httpx_client
```
willbakst: nit: weird to assign the return but return `None`, meaning we should be able to just ignore the return value (or at least make it `_`).

Also, does the client not have a close method? Seems like we should just introduce `.close` if we haven't already.

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,234 @@
...
+    http_client.close()
+
+
+async def _close_async_client(client: AsyncLilypad) -> None:
```
willbakst: ditto

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,234 @@
...
+    http_client = client._client_wrapper.httpx_client.httpx_client
+    await http_client.aclose()
+
+
+def test_lilypad_requires_api_key() -> None:
```
willbakst: nit: tests are also missing docstrings

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,234 @@
...
+    try:
+        with settings_override(api_key=None):
+            lilypad_client = Lilypad(api_key="token-1", httpx_client=client)
+        assert client.headers["Authorization"] == "Bearer token-1"
+        _close_sync_client(lilypad_client)
+    finally:
+        client.close()
+
+
+def test_get_sync_client_caches_instances() -> None:
```
willbakst, line 69: We should pair sync/async tests so we have sync then async for each thing we're testing. This makes it easier to track parity.

sdks/python/tests/lilypad/_internal/test_closure.py
```diff
@@ -430,6 +433,25 @@ def test_closure_with_properties_fn() -> None:
     _assert_dependencies_equal(closure.dependencies, {})
 
 
+def test_kwarg_rich_fn_handles_variadic_parameters() -> None:
```
willbakst, line 807: nit: missing docstrings on all these tests

sdks/python/tests/lilypad/_internal/test_closure.py
```diff
@@ -430,6 +433,25 @@ def test_closure_with_properties_fn() -> None:
...
+    closure = Closure.from_fn(kwarg_rich_fn)
+    assert closure.code == _expected(kwarg_rich_fn)
+    _assert_dependencies_equal(closure.dependencies, {})
+
+
+def test_annotated_dependency_fn_uses_custom_type() -> None:
+    closure = Closure.from_fn(annotated_dependency_fn)
+    assert "Annotated[ModuleLessDependency" in closure.code
+    assert "AnnotatedDependency" in closure.code
```
willbakst: this doesn't use the `_expected` method

sdks/python/tests/lilypad/_internal/test_closure.py
```diff
@@ -605,6 +627,23 @@ def dummy_func(x):
...
+    result.stdout = ""
+    result.stderr = "error"
+
+    with patch("lilypad._internal.closure.subprocess.run", return_value=result):
+
+        def transient() -> str:
+            return "ok"
+
+        Closure.from_fn.cache_clear()
+        with pytest.raises(subprocess.CalledProcessError):
```
willbakst: Do we properly handle this potential error in `VerionedFunction`? Worth making sure this doesn't break user code but logs an error in the event of failure to version.

sdks/python/tests/lilypad/_internal/test_configuration.py
```diff
@@ -127,3 +134,87 @@ def test_configure_without_rich_handler() -> None:
     logger = logging.getLogger("lilypad")
     assert len(logger.handlers) == 1
     assert custom_handler in logger.handlers
+
+
+def test_configure_adds_processors_to_existing_provider() -> None:
```
willbakst, line 170: nit: missing docstrings

sdks/python/tests/lilypad/_internal/test_configuration.py
```diff
@@ -5,9 +5,16 @@
...
+from opentelemetry.trace import INVALID_SPAN_ID, INVALID_TRACE_ID, ProxyTracerProvider
 
+from lilypad._internal.client import close_cached_clients
 from lilypad._internal.configuration import CryptoIdGenerator, configure
+from lilypad._internal.otel.exporters.constants import (
+    DEFAULT_MAX_RETRY_ATTEMPTS,
+    DEFAULT_TIMEOUT,
+)
+from lilypad._internal.otel.exporters.processors import LilypadSpanProcessor
+from lilypad._internal.settings import settings as settings_override
```
willbakst: Why do we alias this?

sdks/python/tests/lilypad/_internal/test_configuration.py
```diff
@@ -127,3 +134,87 @@ def test_configure_without_rich_handler() -> None:
...
+            if isinstance(processor, LilypadSpanProcessor):
+                processor.shutdown()
+    finally:
+        trace._TRACER_PROVIDER = original_provider
+        close_cached_clients()
+
+
+def test_configure_initializes_provider_from_proxy() -> None:
+    original_provider = trace._TRACER_PROVIDER
+    once = getattr(trace, "_TRACER_PROVIDER_SET_ONCE", None)
```
willbakst: I don't really understand what's going on with this `once` stuff. I'll need to take a deeper look.

sdks/python/tests/lilypad/_internal/test_configuration.py
```diff
@@ -127,3 +134,87 @@ def test_configure_without_rich_handler() -> None:
...
+    trace._TRACER_PROVIDER = ProxyTracerProvider()
+    if once is not None:
+        once._done = False
+    try:
+        with settings_override(
+            api_key="token-config-3", base_url="https://example.org"
+        ):
+            configure(timeout=0.0, max_retry_attempts=0)
+        provider = trace.get_tracer_provider()
+        processors = getattr(provider._active_span_processor, "_span_processors", [])  # pyright: ignore[reportAttributeAccessIssue]
```
willbakst: we shouldn't have any ignores unless absolutely necessary.

sdks/python/tests/lilypad/_internal/test_errors.py
```diff
@@ -0,0 +1,12 @@
+"""Tests for custom exceptions."""
+
+from __future__ import annotations
+
+from lilypad._internal.errors import NotFoundError
+
+
+def test_not_found_error_initializes_with_status_code() -> None:
```
willbakst: I would rather think through how to get coverage for `NotFoundError` at a higher e2e level since a user would never really use it like this but may receive it.

sdks/python/tests/lilypad/_internal/test_settings.py
```diff
@@ -0,0 +1,55 @@
...
+
+from lilypad._internal.settings import (
+    CURRENT_SETTINGS,
+    Settings,
+    get_settings,
+    settings,
+)
+
+
+def test_settings_update_skips_none_values() -> None:
```
willbakst, line 8: nit: no docstrings

sdks/python/tests/lilypad/_internal/test_settings.py
```diff
@@ -0,0 +1,55 @@
...
+    instance = Settings(base_url="https://base", api_key="initial")
```
willbakst: Similarly, I don't think a user will ever do this. I think this provides no additional coverage given user-style interface tests below.

sdks/python/tests/lilypad/_internal/test_settings.py
```diff
@@ -0,0 +1,55 @@
...
+    with settings(api_key="ctx-key", base_url="https://ctx.example"):
+        scoped = get_settings()
+        assert scoped.api_key == "ctx-key"
+        assert scoped.base_url == "https://ctx.example"
+    restored = get_settings()
+    assert restored.api_key == original_api_key
+    assert restored.base_url == original_base_url
+
+
+def test_settings_nested_context_restores_previous() -> None:
```
willbakst, line 20: This seems to have duplicate coverage as the above test

sdks/python/tests/lilypad/_internal/test_settings.py
```diff
@@ -0,0 +1,55 @@
...
+    assert restored.api_key == original.api_key
+    assert restored.base_url == original.base_url
+
+
+def test_settings_context_initializes_from_default() -> None:
+    token = CURRENT_SETTINGS.set(None)
+    try:
+        with settings(api_key="fresh-key", base_url="https://fresh.example") as scoped:
+            assert scoped.api_key == "fresh-key"
+            assert scoped.base_url == "https://fresh.example"
```
willbakst: We should also assert that it's the value for `CURRENT_SETTINGS`

sdks/python/tests/lilypad/_internal/test_span.py
```diff
@@ -233,3 +239,114 @@ def test_span_concurrent_operations(span_exporter: InMemorySpanExporter) -> None
             "events": [{"name": "still-active", "attributes": {}}],
         }
     )
+
+
+def test_noop_span_logs_once() -> None:
```
willbakst: nit: no docstrings

sdks/python/tests/lilypad/_internal/test_span.py
```diff
@@ -233,3 +239,114 @@ def test_span_concurrent_operations(span_exporter: InMemorySpanExporter) -> None
...
+            attributes: dict[str, object] | None = None,
+        ):
+            return _StaticNoOpTracer()
+
+    try:
+        otel_trace._TRACER_PROVIDER = _StaticNoOpProvider()
+        from lilypad._internal import spans as span_module
+
+        span_module._warned_noop = False
+        with lilypad.span("noop-log") as first:
```
willbakst, line 296: I don't really see how this test that it logs once?

sdks/python/tests/lilypad/_internal/test_span.py
```diff
@@ -233,3 +239,114 @@ def test_span_concurrent_operations(span_exporter: InMemorySpanExporter) -> None
...
+    span_exporter: InMemorySpanExporter,
+) -> None:
+    with lilypad.span("complex") as span:
+        span.set(numbers=[1, 2, 3], payload='{"ok": true}')
+        span.event("details", points=[1, 2], info='{"nested": "yes"}')
+    spans = span_exporter.get_finished_spans()
+    assert len(spans) == 1
+    data = extract_span_data(spans[0])
+
+    normalized = {
```
willbakst: why don't we just snapshot against data?

sdks/python/tests/lilypad/_internal/test_span.py
```diff
@@ -233,3 +239,114 @@ def test_span_concurrent_operations(span_exporter: InMemorySpanExporter) -> None
...
+def test_span_trace_and_context_properties(span_exporter: InMemorySpanExporter) -> None:
+    with lilypad.span("context-check") as span:
+        trace_id = span.trace_id
+        span_id = span.span_id
+        context = span.span_context
+    spans = span_exporter.get_finished_spans()
+    assert len(spans) == 1
+    otel_context = spans[0].get_span_context()
+    assert otel_context is not None
+    assert trace_id == format(otel_context.trace_id, "032x")
```
willbakst: we should use our existing format methods for these

sdks/python/tests/lilypad/_internal/test_tracing.py
```diff
@@ -157,6 +159,22 @@ def failing_function() -> None:
     assert any(e.name == "exception" for e in events)
 
 
+def test_trace_records_annotation_name(span_exporter: InMemorySpanExporter):
```
willbakst: nit: missing docstrings

sdks/python/tests/lilypad/_internal/test_tracing.py
```diff
@@ -157,6 +159,22 @@ def failing_function() -> None:
...
+    @trace()
+    def typed(value: int) -> int:
+        return value
+
+    typed(5)
+
+    spans = span_exporter.get_finished_spans()
```
willbakst: It seems like this test would be covered by any other test that runs a traced function. We may as well just have a single test that runs a traced function and snapshots against the entire span data.

sdks/python/tests/lilypad/_internal/test_tracing.py
```diff
@@ -182,6 +200,24 @@ def local_func():
         assert "local_func" in qualname_attr
 
 
+def test_trace_handles_runtime_annotations(span_exporter: InMemorySpanExporter):
+    def dynamic(value):
+        return value
+
+    dynamic.__annotations__ = {"value": int, "return": int}
```
willbakst: Nothing tests the return type annotation provided here, so why include it?

sdks/python/tests/lilypad/_internal/test_tracing.py
```diff
@@ -230,3 +266,100 @@ async def async_multiply(x: int, y: int) -> int:
...
+    def optional_output(x: int) -> None:
+        if x > 0:
+            return None
+        return None
+
+    result = optional_output.wrapped(1)
+    assert result.span_id is not None
+    assert result.trace_id is not None
+
+    spans = span_exporter.get_finished_spans()
```
willbakst, line 122: This seems to be overlapping two different tests. The first is testing `.wrapped` as a method, whereas this is testing that we exclude "lilypad.trace.output" when there is no return as omission is the indicator of a `None` return type.

sdks/python/tests/lilypad/_internal/test_tracing.py
```diff
@@ -230,3 +266,100 @@ async def async_multiply(x: int, y: int) -> int:
...
+
+
+def test_trace_handles_unserializable_result(
+    span_exporter: InMemorySpanExporter,
+) -> None:
+    class CustomObject:
+        def __init__(self, value: str) -> None:
+            self.value = value
+
+        def __str__(self) -> str:
```
willbakst: What if it doesn't have a `__str__` method?

sdks/python/tests/lilypad/_internal/test_tracing.py
```diff
@@ -230,3 +266,100 @@ async def async_multiply(x: int, y: int) -> int:
...
+    def produce() -> CustomObject:
+        return CustomObject("value")
+
+    result = produce()
+    assert isinstance(result, CustomObject)
+
+    spans = span_exporter.get_finished_spans()
+    assert len(spans) == 1
+    attributes = spans[0].attributes or {}
+    assert attributes.get("lilypad.trace.output") == "custom:value"
```
willbakst: Shouldn't we be storing the `repr` here so that we could possibly reconstruct the object? My understanding is that `repr` should produce a string from which you can properly reconstruct the given object?

sdks/python/tests/lilypad/_internal/test_tracing.py
```diff
@@ -230,3 +266,100 @@ async def async_multiply(x: int, y: int) -> int:
...
+    def combine(value, count: int) -> str:
+        return f"{value}-{count}"
+
+    combine("name", 3)
+    spans = span_exporter.get_finished_spans()
+    assert len(spans) == 1
+    attributes = spans[0].attributes or {}
+    arg_types = attributes.get("lilypad.trace.arg_types")
+    assert arg_types is not None
+    assert '"value"' in arg_types  # pyright: ignore[reportOperatorIssue]
```
willbakst: no ignores

sdks/python/tests/lilypad/_internal/test_tracing.py
```diff
@@ -230,3 +266,100 @@ async def async_multiply(x: int, y: int) -> int:
...
+
+
+def _make_annotated_function() -> Callable[[int, int], int]:
```
willbakst: I don't understand the need to execute the function into the namespace rather than just using some defined function.

sdks/python/tests/lilypad/_internal/test_tracing.py
```diff
@@ -230,3 +266,100 @@ async def async_multiply(x: int, y: int) -> int:
...
+    span_exporter: InMemorySpanExporter,
+) -> None:
+    traced = trace(_make_annotated_function())
+    result = traced(2, 3)
+    assert result == 5
+    spans = span_exporter.get_finished_spans()
+    assert len(spans) == 1
+
+
+def test_trace_returns_decorator_when_no_function() -> None:
```
willbakst: Better to test that `tags` work properly?

sdks/python/tests/lilypad/_internal/test_tracing.py
```diff
@@ -230,3 +266,100 @@ async def async_multiply(x: int, y: int) -> int:
...
+    decorator = trace(tags=["b", "a", "b"])
+    assert isinstance(decorator, TraceDecorator)
+
+
+@pytest.mark.asyncio
+async def test_trace_direct_call_on_async_function(
+    span_exporter: InMemorySpanExporter,
+) -> None:
+    async def perform(value: int) -> int:
```
willbakst: Yeah case and point right here that the inline defined function tests the same thing as I thought (unless I'm missing something)

sdks/python/tests/lilypad/_internal/test_utils.py
```diff
@@ -28,3 +28,25 @@ def test_json_dumps_uses_correct_orjson_options(mock_dumps: Mock) -> None:
...
+    def inner_function() -> str:
+        return "ok"
+
+    return inner_function
+
+
+def sample_module_level_function() -> None: ...
+
+
+def test_get_qualified_name_for_local_function() -> None:
```
willbakst, line 40: nit: no docstrings

sdks/python/tests/lilypad/_internal/test_utils.py
```diff
@@ -28,3 +28,25 @@ def test_json_dumps_uses_correct_orjson_options(mock_dumps: Mock) -> None:
...
+    inner = _create_inner_function()
+    assert get_qualified_name(inner) == "inner_function"
+
+
+def test_get_qualified_name_for_module_function() -> None:
+    assert (
+        get_qualified_name(sample_module_level_function)
```
willbakst: shouldn't this also just work with `"_create_inner_function"`?

sdks/python/tests/lilypad/_internal/closure_test_functions/main.py
```diff
@@ -679,6 +693,44 @@ def closure_with_properties_fn() -> str:
...
+    """
+    def kwarg_rich_fn(*values: int, keyword: int, **extras: int) -> int:
+        global UNUSED_GLOBAL
+        total = sum(values) + keyword + sum(extras.values())
+        UNUSED_GLOBAL = "updated"
+        return total
+    """
+    global UNUSED_GLOBAL
+    total = sum(values) + keyword + sum(extras.values())
+    UNUSED_GLOBAL = "updated"
```
willbakst: This says unused, but it appear to be used right here?

sdks/python/tests/lilypad/_internal/closure_test_functions/main.py
```diff
@@ -679,6 +693,44 @@ def closure_with_properties_fn() -> str:
     return model.foo + model.bar
 
 
+def kwarg_rich_fn(*values: int, keyword: int, **extras: int) -> int:
+    """
```
willbakst: I'm thinking we could greatly improve these tests by using snapshots instead of docstrings and utilities for testing things.

sdks/python/tests/lilypad/_internal/test_versioning.py
```diff
@@ -3,13 +3,22 @@
...
+import lilypad._internal.versioning as versioning
+from lilypad._generated.types.function import Function
+from lilypad._internal.versioning import (
+    AsyncVersionedFunction,
+    VersionDecorator,
+    VersionedFunction,
+    version,
+)
 
 from .._internal.otel.test_utils import extract_span_data
```
willbakst, line 19: <p class='graphite__hidden'><i>[Re: line +27]</i></p>
We don't need to mock this anymore given it's implemented.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/705?utm_source=unchanged-line-comment">Graphite</a>.</p>

sdks/python/tests/lilypad/_internal/test_versioning.py
```diff
@@ -323,3 +332,433 @@ async def process_data(data: dict[str, int]) -> dict[str, float]:
...
+            )
+        )
+        if self.create_error is not None:
+            raise self.create_error
+        if self.create_result is None:
+            raise RuntimeError("missing create result")
+        return self.create_result
+
+
+class FakeAsyncClient:
```
willbakst: We should avoid dummy classes like this and just use mocks if there really isn't a better way. This is definitely worse.

sdks/python/tests/lilypad/_internal/test_versioning.py
```diff
@@ -323,3 +332,433 @@ async def process_data(data: dict[str, int]) -> dict[str, float]:
...
+    def __init__(self, functions: FakeAsyncFunctions) -> None:
+        self.functions = functions
+
+
+def test_version_direct_sync_returns_wrapper() -> None:
```
willbakst: I'm going to audit all of these tests more thoroughly.

sdks/python/src/lilypad/_internal/closure.py
koxudaxi: Are these comments for all functions, not just this one?

koxudaxi: I made mistake. I should reply the comment to another your comment.

sdks/python/tests/lilypad/_internal/closure_test_functions/main.py
koxudaxi: Are these comments for all functions, not just this one?

koxudaxi: OK, I just applied the changes to all existing closure tests.

sdks/python/tests/lilypad/_internal/closure_test_functions/main.py
```diff
@@ -66,6 +68,47 @@ def wrapper(*args, **kwargs):
...
+
+GLOBAL_STATE = "skip"
+
+
+class ModuleLessDependency:
+    def __init__(self, value: str) -> None:
+        self.value = value
+
+
+ModuleLessDependency.__module__ = ""
```
willbakst: Can we not just do something like this?

```python
class ModuleLessDependency:
    __module__ = ""

    def __init__(self, value: str) -> None:
        self.value = value
```

sdks/python/tests/lilypad/_internal/closure_test_functions/other.py
```diff
@@ -9,6 +9,9 @@
 from lilypad._internal.closure import Closure
 
 
+SELF_REF = "other-module"
```
willbakst: What makes this a self ref?

sdks/python/tests/lilypad/_internal/otel/exporters/test_exporters.py
```diff
@@ -645,220 +662,516 @@ def test_full_pipeline_with_vcr(
         )
     )
 
-    assert all_calls_data == snapshot(
+
+def _make_recording_client() -> tuple[MagicMock, list[list[ResourceSpans]]]:
```
willbakst: nit: let's use `create` instead of `make`?

sdks/python/tests/lilypad/_internal/otel/exporters/test_exporters.py
```diff
@@ -645,220 +662,516 @@ def test_full_pipeline_with_vcr(
...
+    span.instrumentation_scope = instrumentation or InstrumentationScope(
+        name="dummy", version="1.0.0"
+    )
+    span.parent = None if parent_span_id is None else Mock(span_id=parent_span_id)
+    span.events = []
+    span.links = []
+    return span
+
+
+def _build_context(trace_id: int, span_id: int, *, is_valid: bool) -> SpanContext:
```
willbakst: nit: missing docstring (as well as many other methods below, please fix all)

sdks/python/tests/lilypad/_internal/otel/exporters/test_exporters.py
```diff
@@ -645,220 +662,516 @@ def test_full_pipeline_with_vcr(
...
+                    is_valid=True,
+                )
+            ],
+            attributes={},
+            resource_attrs={},
+        ),
+    )
+
+
+def test_exporter_requires_traces_interface() -> None:
```
willbakst: none of these tests have docstrings

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,241 @@
...
+def reset_client_caches() -> Generator[None, None, None]:
+    """Ensure cached clients are cleared before and after each test."""
+    close_cached_clients()
+    yield
+    close_cached_clients()
+
+
+def _close_sync_client(client: Lilypad) -> None:
+    """Close the underlying HTTPX client for a synchronous Lilypad instance."""
+    client._client_wrapper.httpx_client.httpx_client.close()
```
willbakst: Why don't we introduce a `.close` method on `Lilypad`?

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+async def _close_async_client(client: AsyncLilypad) -> None:
+    """Close the underlying HTTPX client for an asynchronous Lilypad instance."""
+    await client._client_wrapper.httpx_client.httpx_client.aclose()
```
willbakst: ditto

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,241 @@
...
+        first = get_async_client()
+        second = get_async_client()
+    assert first is second
+    await _close_async_client(first)
+
+
+def test_get_sync_client_requires_api_key() -> None:
+    """Raise when requesting a synchronous client without an API key."""
+    with settings(api_key=None), pytest.raises(RuntimeError):
+        get_sync_client()
```
willbakst, line 93: Why no async equivalent?

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+def test_get_sync_client_handles_invalid_cached_instance() -> None:
+    """Validate inputs when requesting cached sync clients."""
+    with settings(api_key=""), pytest.raises(RuntimeError):
+        get_sync_client(api_key="")
+
+
+@pytest.mark.asyncio
+async def test_get_async_client_handles_invalid_key() -> None:
```
willbakst: Hmm, seems like this is the matching test for above? The tests need to properly match and be paired up unless there is only one test to write (e.g. async event loop test), which should probably go at the end as the extra async edge case tests

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,241 @@
...
+    await _close_async_client(first)
+    close_cached_clients()
+    with settings(api_key="token-7"):
+        second = get_async_client()
+    assert id(second) != first_id
+    await _close_async_client(second)
+
+
+@pytest.mark.asyncio
+async def test_async_lilypad_default_httpx_configuration() -> None:
```
willbakst, line 173: where is the sync test for this? If we don't need it, do we need this test?

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,241 @@
...
+    finally:
+        await _close_async_client(client)
+
+
+def test_get_sync_client_handles_settings_failure() -> None:
+    """Surface failures encountered while resolving synchronous settings."""
+
+    with (
+        patch(
+            "lilypad._internal.client.get_settings",
```
willbakst: When would this happen without mocking our internal implementation?

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,241 @@
...
+            side_effect=RuntimeError("broken settings"),
+        ),
+        pytest.raises(RuntimeError, match="broken settings"),
+    ):
+        get_async_client(api_key="token")
+    close_cached_clients()
+
+
+def test_get_sync_client_propagates_initialization_error() -> None:
```
willbakst: Why do we need to test that an error will propagate? Isn't it usually the opposite? That we test we handle an error?

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,241 @@
...
+            pytest.raises(RuntimeError),
+        ):
+            get_sync_client()
+    finally:
+        client_module.Lilypad = original_class  # type: ignore[assignment]
+        close_cached_clients()
+
+
+@pytest.mark.asyncio
+async def test_get_async_client_propagates_initialization_error() -> None:
```
willbakst: ditto

sdks/python/src/lilypad/_internal/versioning.py
```diff
@@ -65,18 +66,26 @@ class AsyncVersionedResult(AsyncTracedResult[R]):
...
 
-    closure: Closure = field(init=False)
+    closure: Closure | None = field(init=False, default=None)
 
     def __post_init__(self) -> None:
         super().__post_init__()
-        self.closure = Closure.from_fn(self.fn)
+        try:
+            self.closure = Closure.from_fn(self.fn)
+        except subprocess.CalledProcessError as error:
```
willbakst: The `Closure.from_fn` method should raise an error rather than let the ruff formatting error propagate.

For example:

```python
Raises:
    ClosureFormattingError: If `ruff` fails to run and the code cannot be
        formatted
```

Then it will be more clear when using `Closure.from_fn` that you need to handle said failure.

sdks/python/tests/lilypad/_internal/test_utils.py
willbakst: We can clean this up by removing this since we test `_create_inner_function` below.

sdks/python/tests/lilypad/_internal/test_tracing.py
willbakst, line 1: This can be refactored like `test_versioning.py` was.

sdks/python/tests/lilypad/_internal/test_span.py
```diff
@@ -2,22 +2,74 @@
...
+        name: str,
+        version: str | None = None,
+        schema_url: str | None = None,
+        attributes: dict[str, object] | None = None,
+    ):
+        return _StaticNoOpTracer()
+
+
+@contextmanager
+def _use_noop_provider():
```
willbakst: This and the `noop_provider` fixtures are essentially duplicated?

sdks/python/tests/lilypad/_internal/test_span.py
```diff
@@ -2,22 +2,74 @@
...
+    """Fixture that provides a no-op tracer provider for testing."""
+    original_provider = otel_trace._TRACER_PROVIDER
+    otel_trace._TRACER_PROVIDER = _StaticNoOpProvider()
+    try:
+        yield
+    finally:
+        otel_trace._TRACER_PROVIDER = original_provider
+
+
+def test_noop_span_basic(noop_provider) -> None:
```
willbakst: nit: missing type hint

sdks/python/tests/lilypad/_internal/test_settings.py
```diff
@@ -0,0 +1,55 @@
...
+
+def test_settings_update_skips_none_values() -> None:
+    """Update settings values while ignoring fields explicitly set to None."""
+    with settings(api_key="initial", base_url="https://base") as scoped:
+        scoped.update(api_key="updated", base_url=None)
+        assert scoped.api_key == "updated"
+        assert scoped.base_url == "https://base"
+
+
+def test_settings_context_overrides_and_restores() -> None:
```
willbakst: Let's remove this in favor of the test below, which covers this.

sdks/python/tests/lilypad/_internal/test_settings.py
```diff
@@ -0,0 +1,55 @@
...
+        assert scoped.base_url == "https://ctx.example"
+    restored = get_settings()
+    assert restored.api_key == original_api_key
+    assert restored.base_url == original_base_url
+
+
+def test_settings_nested_context_restores_previous() -> None:
+    """Restore outer settings after exiting a nested override."""
+    original = get_settings()
+    with settings(api_key="outer-key", base_url="https://outer.example"):
```
willbakst: let's refactor all of these strings into variables

sdks/python/tests/lilypad/_internal/test_configuration.py
```diff
@@ -1,27 +1,73 @@
...
+    def do_once(self, func) -> bool:
+        if self._done:
+            return False
+        func()
+        self._done = True
+        return True
+
+
+@contextmanager
+def patched_provider(target_provider) -> Generator[TracerProvider, None, None]:
```
willbakst: nit: missing type hint

sdks/python/tests/lilypad/_internal/test_configuration.py
```diff
@@ -1,27 +1,73 @@
...
+    """Temporarily patch the global tracer provider and reset the once guard."""
+    with (
+        patch.object(trace, "_TRACER_PROVIDER", target_provider),
+        patch.object(trace, "_TRACER_PROVIDER_SET_ONCE", _ResettableOnce()),
```
willbakst: I see now. Do they not already have something we could use instead of making our own `_ResettableOnce()` object here?

sdks/python/tests/lilypad/_internal/test_closure.py
```diff
@@ -526,7 +1189,67 @@ def test_nested_base_model_definitions() -> None:
...
+    return {"computed_fields": {"ticket": ticket}}
+
+
+def handle_issue(issue: str) -> str:
+    history = []
+    response = customer_support_bot(issue, history)
+    return str(response)
+"""
+    )
     _assert_dependencies_equal(
```
willbakst: Let's also update these across the board to use snapshots

sdks/python/tests/lilypad/_internal/test_closure.py
```diff
@@ -426,15 +869,158 @@ def test_closure_inside_imported_decorator_fn() -> None:
...
+    )
+    _assert_dependencies_equal(closure.dependencies, {})
+
+
+def test_annotated_container_fn_collects_types() -> None:
+    """Ensure class annotations using Annotated are preserved."""
+    Closure.from_fn.cache_clear()
+    closure = Closure.from_fn(annotated_container_fn)
+    assert closure.code == snapshot(
+        inspect.cleandoc(
```
willbakst: Why do some of these have `inspect.cleandoc(...)` while others don't? Let's be consistent (and not use it)?

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,255 @@
...
+
+    with settings(api_key="token-4"):
+        second = get_sync_client()
+
+    assert id(second) != first_id
+    _close_sync_client(second)
+
+
+@pytest.mark.asyncio
+async def test_close_cached_async_clients_resets_singleton() -> None:
```
willbakst: nit: we should rename this and the matching sync case to put sync/async right before singleton

sdks/python/tests/lilypad/_internal/closure_test_functions/main.py
```diff
@@ -679,6 +722,56 @@ def closure_with_properties_fn() -> str:
...
+    return inner()
+
+
+def annotated_container_fn() -> AnnotatedContainer:
+    container = AnnotatedContainer(ModuleLessDependency("value"))
+    return container
+
+
+def annotated_dependency_fn(payload: AnnotatedDependency) -> ModuleLessDependency:
+    """
```
willbakst: We can now remove all of these docstrings since we don't use them for testing anymore.

sdks/python/tests/lilypad/_internal/otel/exporters/test_exporters.py
```diff
@@ -645,220 +662,516 @@ def test_full_pipeline_with_vcr(
...
     )
 
+    provider.shutdown()
+
+
+def test_exporter_converts_unhandled_attribute_to_string() -> None:
+    client, recorded_requests = _make_recording_client()
+    exporter = LilypadOTLPExporter(client=cast(Lilypad, client))
+
+    span = _make_span_mock(
```
willbakst: I still don't understand why we need to fully create this? Can we not just export it and assert that this mock was in fact exported? What part of the specific fields do we need for this to work?

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,263 @@
...
+            lilypad_client = Lilypad(api_key="token-1", httpx_client=client)
+
+        assert client.headers["Authorization"] == "Bearer token-1"
+        lilypad_client.close()
+    finally:
+        client.close()
+
+
+@pytest.mark.asyncio
+async def test_async_lilypad_updates_headers() -> None:
```
willbakst: nit: test name should better match sync case above

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+@pytest.mark.asyncio
+async def test_get_async_client_handles_invalid_cached_instance() -> None:
+    """Test that get_async_client validates inputs when requesting cached clients."""
+    with settings(api_key=""), pytest.raises(RuntimeError):
+        get_async_client(api_key="")
+
+
+def test_create_export_client_uses_configuration() -> None:
```
willbakst, line 116: I don't see how this tests what it says it tests?

sdks/python/tests/lilypad/_internal/test_client.py
```diff
@@ -0,0 +1,263 @@
...
+            side_effect=RuntimeError("broken settings"),
+        ),
+        pytest.raises(RuntimeError, match="broken settings"),
+    ):
+        get_async_client(api_key="token")
+
+    close_cached_clients()
+
+
+def test_get_sync_client_handles_initialization_error() -> None:
```
willbakst: When would this ever happen?

sdks/python/tests/lilypad/_internal/test_closure.py
```diff
@@ -92,81 +96,150 @@ def _assert_dependencies_equal(
...
-            def single_fn() -> str:
-                return "Hello, world!"
-            """
-            return "Hello, world!"
-        ''')
-        + "\n"
+    assert closure.code == snapshot(
+        '''\
+def single_fn() -> str:
+    """
```
willbakst: If we're going to keep a docstring on this function for testing, let's make it a normal docstring now that we're using snapshots and don't need the docstring to look like this anymore.

sdks/python/tests/lilypad/_internal/test_configuration.py
```diff
@@ -1,27 +1,62 @@
...
 def clear_logger_handlers() -> Generator[None, None, None]:
-    """Clear logger handlers after each test to prevent test pollution."""
+    """Clear logger handlers after each test to prevent pollution."""
     yield
     logger = logging.getLogger("lilypad")
     logger.handlers.clear()
     logger.setLevel(logging.WARNING)
 
 
+def _shutdown_span_processors(provider: Any) -> None:
```
willbakst: Why `Any`? Why not a specific type here?

sdks/python/tests/lilypad/_internal/test_tracing.py
```diff
@@ -3,51 +3,116 @@
...
-        return x + 1
+@pytest.fixture(autouse=True)
+def clear_trace_state() -> Generator[None, None, None]:
+    """Ensure trace-related state is cleared between tests."""
+    yield
+
 
-    assert f(3) == 4
-    assert calls == [3]
+def run_traced(
```
willbakst: I don't love this function and it's type ignores

sdks/python/tests/lilypad/_internal/test_tracing.py
```diff
@@ -3,51 +3,116 @@
...
+    """Execute a traced function, handling both sync and async cases."""
+    if is_async:
+        return asyncio.run(fn(*args, **kwargs))  # type: ignore[return-value]
+    return fn(*args, **kwargs)  # type: ignore[return-value]
 
+
+def get_single_span_data(span_exporter: InMemorySpanExporter) -> SpanData:
+    """Extract and return span data from a single span for snapshot comparison."""
     spans = span_exporter.get_finished_spans()
     assert len(spans) == 1
```
willbakst: I don't like having asserts in utilities. Each test should assert specifically what it cares about directly.

sdks/python/tests/lilypad/_internal/test_tracing.py
```diff
@@ -3,51 +3,116 @@
...
 
 
-def test_trace_with_no_tags(span_exporter: InMemorySpanExporter):
-    """Test @trace decorator without tags."""
+@pytest.mark.parametrize("is_async", [False, True])
+def test_trace_basic(span_exporter: InMemorySpanExporter, is_async: bool) -> None:
+    """Test @trace decorator on sync/async functions with tags."""
+    calls = []
+
+    if is_async:
```
willbakst: We should always keep sync and async tests separate unless they can both be handled synchronously

willbakst (approved):

PR #703, opened 2025-09-22
feat: Implement Mistral OpenTelemetry

koxudaxi:
### TL;DR

Added support for Mistral AI in the Python SDK's telemetry and tracing system.

### What changed?

- Added Mistral AI as a new provider in the Python SDK's OpenTelemetry instrumentation
- Implemented a `_MistralInstrumentor` class to handle Mistral client instrumentation
- Created the `instrument_mistral()` function to instrument Mistral clients
- Added necessary dependencies in `pyproject.toml` for the new Mistral integration
- Added comprehensive test coverage for both sync and async Mistral client operations
- Updated the OpenTelemetry initialization to include Mistral instrumentation

### How to test?

1. Install the Mistral dependency:
   ```
   pip install lilypad[mistral]
   ```

2. Use the instrumentation in your code:
   ```python
   from mistralai import Mistral
   from lilypad._internal.otel import instrument_mistral

   client = Mistral(api_key="your-api-key")
   instrument_mistral(client)

   # Now use the client as normal, with telemetry enabled
   response = client.chat.complete(
       model="mistral-small-latest",
       messages=[{"role": "user", "content": "Hello world"}]
   )
   ```

3. Run the tests:
   ```
   cd sdks/python
   pytest tests/lilypad/_internal/otel/test_mistral.py
   ```

### Why make this change?

This change extends the SDK's telemetry capabilities to include Mistral AI, allowing users to collect standardized telemetry data when using Mistral models. This is part of the ongoing effort to provide comprehensive observability across all supported LLM providers, making it easier for users to monitor, debug, and analyze their AI applications regardless of which provider they're using.

PR #706, opened 2025-10-06
Implement context propagation in Python SDK

koxudaxi:
# Add distributed tracing context propagation for HTTP clients

### TL;DR

Add support for propagating trace context across service boundaries via HTTP clients with instrumentation for httpx, requests, aiohttp, and urllib3.

### What changed?

- Added OpenTelemetry propagators for B3 and Jaeger formats
- Created a new context propagation system that supports multiple propagation formats
- Implemented HTTP client instrumentation for:
  - httpx (sync and async)
  - requests
  - aiohttp
  - urllib3
- Added `propagated_context()` context manager for server-side context extraction
- Added comprehensive test coverage for all new functionality

### How to test?

```python
from lilypad import configure, instrument_httpx, trace

# Initialize tracing
configure()

# Instrument HTTP client
instrument_httpx()

# Create a span that will be propagated in outgoing requests
with trace("parent_operation"):
    # Any HTTP requests made here will automatically include trace context
    response = httpx.get("https://api.example.com/endpoint")
```

For server-side context extraction:

```python
from lilypad import propagated_context

@app.post("/endpoint")
async def endpoint(request: Request):
    # Extract and attach trace context from incoming request headers
    with propagated_context(extract_from=dict(request.headers)):
        result = process_request()
    return result
```

### Why make this change?

This change enables distributed tracing across service boundaries, allowing users to track requests as they flow through multiple services. By automatically propagating trace context in HTTP requests, users can create end-to-end traces without manually managing context propagation. This is essential for observability in microservice architectures and distributed systems.


koxudaxi:
* **#708** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/708?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#707** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/707?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#706** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/706?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/706?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `v1`


<h2></h2>

sdks/python/pyproject.toml
```diff
@@ -41,15 +43,19 @@ anthropic = ["anthropic>=0.63.0"]
 [dependency-groups]
 dev = ["pyright>=1.1.403", "pre-commit>=3.5.0"]
 test = [
+    "aiohttp>=3.11.11",
+    "anthropic>=0.63.0",
```
willbakst: ??

sdks/python/src/lilypad/__init__.py
```diff
@@ -24,6 +29,18 @@
 with suppress(ImportError):
     from ._internal.otel import instrument_anthropic
 
+with suppress(ImportError):
+    from ._internal.otel.http import instrument_requests
+    from ._internal.otel.http import uninstrument_requests
```
willbakst, line 34: Should we also have "uninstrument" methods for OpenAI and Anthropic?

sdks/python/src/lilypad/_internal/otel/http/_base_instrumentor.py
```diff
@@ -0,0 +1,166 @@
...
+    """Whether the method is async (True) or sync (False)."""
+
+    error_log_level: int
+    """The logging level to use when context injection fails (e.g., logging.WARNING)."""
+
+    error_log_message_template: str
+    """Template for error log message. Should contain '{exception_type}' and '{exception}' placeholders."""
+
+
+class HttpContextInstrumentor(BaseInstrumentor, ABC):
```
willbakst, line 46: This looks clean. Are there improvements you've made here that you think we should apply to our Gen AI instrumentation?

sdks/python/src/lilypad/_internal/otel/propagation.py
```diff
@@ -0,0 +1,177 @@
...
+    """Manages OpenTelemetry context propagation across service boundaries."""
+
+    def __init__(self, *, set_global: bool = True) -> None:
+        """Initialize propagator and optionally set as global.
+
+        Propagator format is determined by LILYPAD_PROPAGATOR environment variable.
+        Defaults to "tracecontext" if not set.
+
+        Args:
+            set_global: Whether to set this propagator as the global textmap propagator.
```
willbakst, line 43: nit: why not just `global: bool = True`?

sdks/python/src/lilypad/_internal/otel/http/_base_instrumentor.py
```diff
@@ -0,0 +1,166 @@
...
+        headers = kwargs.get("headers")
+        if headers is None:
+            headers = {}
+            kwargs["headers"] = headers
+        elif not isinstance(headers, dict):
+            headers = dict(headers)
+            kwargs["headers"] = headers
+
+        try:
+            inject_context(headers)
```
willbakst: Can you explain why we use the global method here rather than maintaining an internal propagator?

sdks/python/tests/lilypad/_internal/otel/http/conftest.py
```diff
@@ -0,0 +1,76 @@
...
+    provider = TracerProvider()
+    trace.set_tracer_provider(provider)
+
+    yield provider
+
+    trace.set_tracer_provider(original_provider)
+
+
+@pytest.fixture
+def header_capture_handler(
```
willbakst: This one doesn't end with `_fixture`. I think we should either have them all end with `fixture` or drop `fixture`. Which would you prefer?

sdks/python/tests/lilypad/_internal/otel/http/test_aiohttp.py
```diff
@@ -0,0 +1,229 @@
...
+            ) as response:
+                assert response.status == 200
+
+    normalized = normalize_headers(captured_headers)
+    assert "Traceparent" in normalized
+    assert normalized["Traceparent"] == "<traceparent>"
+
+
+@pytest.mark.asyncio
+async def test_aiohttp_instrumentation(
```
willbakst, line 59: Seems like there's a fair bit of overlap between this test and the above tests? I feel like we could drop the first test entirely? But keep the `None` check test.

sdks/python/tests/lilypad/_internal/otel/http/test_aiohttp.py
```diff
@@ -0,0 +1,229 @@
...
+    tracer = trace.get_tracer(__name__)
+
+    with caplog.at_level(logging.WARNING), tracer.start_as_current_span("test-span"):
+        async with (
+            aiohttp.ClientSession() as session,
+            session.get(httpserver.url_for("/test")) as response,
+        ):
+            assert response.status == 200
+
+        assert "Failed to inject context into aiohttp request" in caplog.text
```
willbakst: Why not just assert against a snapshot of the caplog?

sdks/python/tests/lilypad/_internal/otel/http/test_aiohttp.py
```diff
@@ -0,0 +1,229 @@
...
+    """Return correct instrumentation dependencies."""
+    instrumentor = AiohttpInstrumentor()
+
+    dependencies = instrumentor.instrumentation_dependencies()
+
+    assert dependencies == snapshot(["aiohttp"])
+
+
+@pytest.mark.asyncio
+async def test_aiohttp_integration_trace_continuity(httpserver: HTTPServer) -> None:
```
willbakst: Why not have this continuity test in the earlier test? There seems to be a good amount of overlap here. Also is the `json` handling part of this test or just for coverage? I suggest having that as it's own test with a snapshot on the captured headers (if the test is necessary)

sdks/python/tests/lilypad/_internal/otel/http/test_utils.py
```diff
@@ -0,0 +1,56 @@
+"""Shared helper functions for HTTP instrumentation tests."""
+
+from __future__ import annotations
+
+import re
+from collections.abc import Iterator, Mapping
+
+
+class DictLikeHeaders(Mapping[str, str]):
```
willbakst: Why not just name this `MockDictHeaders`?

sdks/python/tests/lilypad/_internal/otel/http/test_httpx.py
willbakst, line 1: This has a lot of overlap with `test_aiohttp.py`. My feeling is that the remaining files will also have substantial overlap.

It's likely worth thinking how we could unify these tests. One small example that comes to mind is unifying the `captured_headers` concept into a fixture at the very least. The `httpserver` code is extremely repetitive. You could also have `httpserver.url_for("/test")` as a fixture as well.

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+
+    def __iter__(self) -> Iterator[str]:
+        return iter(self._data)
+
+    def __len__(self) -> int:
+        return len(self._data)
+
+
+@pytest.fixture(autouse=True)
+def reset_propagator_fixture() -> Generator[None, None, None]:
```
willbakst: This is duplicated somewhere I saw earlier?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+    """Reset the global propagator singleton between tests."""
+    yield
+    reset_propagator()
+
+
+@pytest.fixture(autouse=True)
+def reset_environment() -> Generator[None, None, None]:
```
willbakst: Seems like this (and maybe a few other methods) should live in a `conftest.py` for the entire `http` module tests?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+    else:
+        os.environ["LILYPAD_PROPAGATOR"] = original_propagator_env
+
+    if original_set_global_env is None:
+        os.environ.pop("_LILYPAD_PROPAGATOR_SET_GLOBAL", None)
+    else:
+        os.environ["_LILYPAD_PROPAGATOR_SET_GLOBAL"] = original_set_global_env
+
+
+def configure_propagator(format_name: str) -> None:
```
willbakst: nit: private method?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+    ("propagator_env", "expected_headers"),
+    [
+        (None, {"traceparent"}),
+        ("tracecontext", {"traceparent"}),
+        ("b3", {"b3"}),
+        ("b3multi", {"x-b3-traceid", "x-b3-spanid", "x-b3-sampled"}),
+        ("jaeger", {"uber-trace-id"}),
+        ("composite", {"traceparent"}),
+    ],
+    ids=["default", "tracecontext", "b3", "b3multi", "jaeger", "composite"],
```
willbakst: nit: maybe use `PROPAGATOR_FORMATS` here?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+
+    context_propagator = ContextPropagator(set_global=False)
+
+    tracer = tracer_provider.get_tracer(__name__)
+    with tracer.start_as_current_span("test-span") as span:
+        current_context = trace.set_span_in_context(span)
+        carrier: dict[str, str] = {}
+        context_propagator.inject_context(carrier, context=current_context)
+
+        assert expected_headers.issubset(set(carrier.keys())), (
```
willbakst: I think you can just substract them and if it's not the empty set then the remaining item would be the items missing, which would give a more granular error message when this assertion fails.

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+    """Raise ConfigurationError and log error for invalid propagator format."""
+    os.environ["LILYPAD_PROPAGATOR"] = "invalid_format"
+
+    with (
+        caplog.at_level(logging.ERROR),
+        pytest.raises(ConfigurationError) as exception_info,
+    ):
+        ContextPropagator()
+
+    assert "Invalid propagator format: invalid_format" in str(exception_info.value)
```
willbakst: Let's snapshot these rather than asserting the strings manually? That will make updating these less annoying if we ever want to.

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+    assert "Invalid propagator format: invalid_format" in caplog.text
+    assert "Valid options:" in caplog.text
+
+
+def test_propagator_sets_global_by_default() -> None:
+    """Set the propagator as global textmap propagator by default."""
+    os.environ["LILYPAD_PROPAGATOR"] = "tracecontext"
+
+    with patch.object(propagate, "set_global_textmap") as mock_set_global:
```
willbakst: Why mock this rather than check if the global propagator is in fact this propagator?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+        ContextPropagator()
+        assert mock_set_global.call_count == 1
+
+
+def test_propagator_respects_set_global_false() -> None:
+    """Do not set global propagator when set_global=False."""
+    os.environ["LILYPAD_PROPAGATOR"] = "tracecontext"
+
+    with patch.object(propagate, "set_global_textmap") as mock_set_global:
```
willbakst: Ditto

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+        ContextPropagator(set_global=False)
+        mock_set_global.assert_not_called()
+
+
+def test_propagator_respects_set_global_guard() -> None:
+    """Do not set global propagator when _LILYPAD_PROPAGATOR_SET_GLOBAL is false."""
+    os.environ["LILYPAD_PROPAGATOR"] = "tracecontext"
+    os.environ["_LILYPAD_PROPAGATOR_SET_GLOBAL"] = "false"
+
+    with patch.object(propagate, "set_global_textmap") as mock_set_global:
```
willbakst: Ditto

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+    """Extract context from carrier with valid traceparent header."""
+    os.environ["LILYPAD_PROPAGATOR"] = "tracecontext"
+
+    context_propagator = ContextPropagator(set_global=False)
+    carrier = {"traceparent": "00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01"}
+
+    extracted_context = context_propagator.extract_context(carrier)
+
+    assert isinstance(extracted_context, Context)
+    assert extracted_context is not None
```
willbakst: Why not snapshot this?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+    """Return empty context when carrier has malformed headers."""
+    os.environ["LILYPAD_PROPAGATOR"] = "tracecontext"
+
+    context_propagator = ContextPropagator(set_global=False)
+    carrier = {"traceparent": "invalid-traceparent-format"}
+
+    with caplog.at_level(logging.DEBUG):
+        extracted_context = context_propagator.extract_context(carrier)
+
+    assert isinstance(extracted_context, Context)
```
willbakst, line 226: How does this test that it's empty?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+def test_propagator_extracts_empty_context_on_invalid_headers(
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    """Return empty context when carrier has malformed headers."""
+    os.environ["LILYPAD_PROPAGATOR"] = "tracecontext"
+
+    context_propagator = ContextPropagator(set_global=False)
+    carrier = {"traceparent": "invalid-traceparent-format"}
+
+    with caplog.at_level(logging.DEBUG):
```
willbakst: Why do this if you do nothing with caplog?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+        extracted_context = context_propagator.extract_context(carrier)
+
+    assert isinstance(extracted_context, Context)
+
+
+def test_propagator_extracts_empty_context_on_missing_headers() -> None:
```
willbakst: It feels like these last few tests could be merged with parameterization?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+    """Log debug message when extraction fails on malformed carrier."""
+    os.environ["LILYPAD_PROPAGATOR"] = "tracecontext"
+
+    context_propagator = ContextPropagator(set_global=False)
+
+    with caplog.at_level(logging.DEBUG):
+        carrier = InvalidCarrier()
+        result = context_propagator.extract_context(carrier)
+
+        assert "Failed to extract context from carrier" in caplog.text
```
willbakst: ditto re: snapshots

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,596 @@
...
+    tracer = tracer_provider.get_tracer(__name__)
+
+    context_propagator = ContextPropagator(set_global=False)
+
+    with tracer.start_as_current_span("test-span") as span:
+        current_context = trace.set_span_in_context(span)
+        carrier: dict[str, str] = {}
+        context_propagator.inject_context(carrier, context=current_context)
+
+        assert "traceparent" in carrier
```
willbakst: Why not snapshot? Same question for a lot of these tests?

sdks/python/tests/lilypad/_internal/test_configuration.py
```diff
@@ -24,10 +26,38 @@
...
     logger = logging.getLogger("lilypad")
+    original_level = logger.level
+    yield
     logger.handlers.clear()
-    logger.setLevel(logging.WARNING)
+    logger.setLevel(original_level)
+
+
+@pytest.fixture(autouse=True)
+def reset_propagator() -> Generator[None, None, None]:
```
willbakst: Yeah this seems duplicated...

sdks/python/tests/lilypad/_internal/test_configuration.py
```diff
@@ -229,3 +259,106 @@ def test_configure_defaults_from_zero_values() -> None:
...
+
+
+def test_configure_with_propagator_tracecontext() -> None:
+    """Set up tracecontext propagator via environment variable."""
+    propagation._PROPAGATOR = None
+    os.environ["LILYPAD_PROPAGATOR"] = "tracecontext"
+
+    result = propagation.get_propagator()
+
+    assert os.environ.get("LILYPAD_PROPAGATOR") == "tracecontext"
```
willbakst: This seems a silly test given you just set it above?

sdks/python/tests/lilypad/_internal/test_configuration.py
```diff
@@ -229,3 +259,106 @@ def test_configure_defaults_from_zero_values() -> None:
...
+    assert propagation._PROPAGATOR is not None
+    assert result is propagation._PROPAGATOR
+
+
+def test_configure_with_propagator_b3() -> None:
```
willbakst: Why not parameterize these tests? So much overlap

sdks/python/tests/lilypad/_internal/test_context.py
```diff
@@ -0,0 +1,214 @@
...
+from opentelemetry.context import Context
+from opentelemetry.sdk.trace import TracerProvider
+
+from lilypad._internal.otel.propagation import reset_propagator
+from lilypad._internal.otel import propagation
+from lilypad._internal.context import propagated_context
+
+
+@pytest.fixture(autouse=True)
+def reset_propagator_fixture() -> Generator[None, None, None]:
```
willbakst: duplicated

sdks/python/src/lilypad/_internal/otel/http/_base_instrumentor.py
koxudaxi: `inject_context` uses the global  OpenTelemetry propagator to write trace headers like traceparent into the provided headers mapping. It updates in  place and returns nothing, so there is no value to capture.

sdks/python/pyproject.toml
koxudaxi: Oh, I made mistake. I just remove the line

sdks/python/src/lilypad/__init__.py
koxudaxi: Anthropic and OpenAI instrumentors do not keep the original methods, so the only way to undo is to create a fresh client. Should we add the same uninstrument flow for these providers? If yes, I would like to handle it in a separate task.

sdks/python/src/lilypad/_internal/otel/http/_base_instrumentor.py
koxudaxi: A possible change would be to add a small spec so we can register each wrapped method.
That way we keep the original bound callable before wrapping it, which is what we need to  enable uninstrument.

sdks/python/src/lilypad/_internal/otel/propagation.py
koxudaxi: `global` is a Python keyword, so it can’t be used as a parameter name.

sdks/python/tests/lilypad/_internal/otel/http/test_aiohttp.py
koxudaxi: OK,  I just removed it

sdks/python/tests/lilypad/_internal/otel/http/test_httpx.py
koxudaxi: OK, I refactor it.

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
koxudaxi: How to use snapshot here?

sdks/python/src/lilypad/__init__.py
willbakst: My thinking is yes, but we can put that on the back burner.

sdks/python/src/lilypad/_internal/otel/http/_base_instrumentor.py
willbakst: Ok, I will include this in the issue tracking this work. We can work on this separately in the future.

sdks/python/src/lilypad/_internal/otel/http/_base_instrumentor.py
willbakst: So the `HttpContextInstrumentor ` always uses the global propagator? Shouldn't it set something like `self.propagator` to the global if not set but otherwise use the given propagator? What am I missing?

sdks/python/tests/lilypad/_internal/conftest.py
willbakst, line 1: nit: shouldn't this go in the conftest.py for the otel module?

sdks/python/tests/lilypad/_internal/otel/http/test_aiohttp.py
```diff
@@ -0,0 +1,179 @@
...
+    header_capture_fixture: HeaderCaptureSetup,
+) -> None:
+    """Verify instrumentation can be enabled and disabled, affecting header injection."""
+    capture = header_capture_fixture("test")
+
+    tracer = trace.get_tracer(__name__)
+
+    with tracer.start_as_current_span("test-span"):
+        async with aiohttp.ClientSession() as session:
+            async with session.get(capture.urls[0]) as response:
```
willbakst, line 69: I'm just now realizing that this tests `.get` but the below tests `.post`?

sdks/python/tests/lilypad/_internal/otel/http/test_requests.py
```diff
@@ -0,0 +1,139 @@
...
+
+from .conftest import HeaderCaptureSetup, SingleHeaderCaptureSetup
+from .test_utils import (
+    MockDictHeaders,
+    normalize_headers,
+    trace_id_from_traceparent,
+)
+
+
+def test_requests_injects_headers(
```
willbakst: Do we need this test given the test below?

sdks/python/tests/lilypad/_internal/otel/http/test_httpx.py
```diff
@@ -0,0 +1,194 @@
...
+
+from .conftest import HeaderCaptureSetup, SingleHeaderCaptureSetup
+from .test_utils import (
+    MockDictHeaders,
+    normalize_headers,
+    trace_id_from_traceparent,
+)
+
+
+def test_httpx_sync_injects_headers(
```
willbakst: Do we need this given the `test_httpx_instrumentation` test below? Seems to duplicate what it's testing

sdks/python/tests/lilypad/_internal/otel/http/test_httpx.py
```diff
@@ -0,0 +1,194 @@
...
+            "Accept-Encoding": "gzip, deflate",
+            "Connection": "keep-alive",
+            "User-Agent": "<user-agent-httpx>",
+            "Traceparent": "<traceparent>",
+        }
+    )
+
+
+@pytest.mark.asyncio
+async def test_httpx_async_injects_headers(
```
willbakst: Same thing here. Shouldn't we have an async test matching the `test_httpx_instrumentation` test below?

sdks/python/tests/lilypad/_internal/otel/http/utils.py
willbakst, line 1: nit: this should just be named `utils.py` since it's not testing anything but instead is providing utilities?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,618 @@
...
+
+
+@pytest.fixture
+def tracer(tracer_provider: TracerProvider) -> trace.Tracer:
+    """Get a tracer from the test provider."""
+    return trace.get_tracer(__name__)
+
+
+@pytest.fixture(autouse=True)
+def restore_propagation_fixture() -> Iterator[None]:
```
willbakst: Why is this one so much more complicated than the `conftest.py` version of restoring the propagators?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
willbakst: Hmm, I mostly just wonder if `is not None` is enough of a test to ensure it extracts context successfully? Shouldn't we be testing something about the extracted context beyond it's type?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,618 @@
...
+    ContextPropagator()
+    assert propagate.get_global_textmap() is previous_global
+
+
+def test_propagator_extracts_context_successfully() -> None:
+    """Extract context from carrier with valid traceparent header."""
+    os.environ[ENV_PROPAGATOR_FORMAT] = "tracecontext"
+
+    context_propagator = ContextPropagator(set_global=False)
+    carrier = {"traceparent": "00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01"}
```
willbakst: Why not use the carrier fixture I've seen elsewhere?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,618 @@
...
+
+@pytest.mark.parametrize(
+    "carrier",
+    [
+        {"traceparent": "invalid-traceparent-format"},
+        {},
+    ],
+    ids=["invalid_headers", "missing_headers"],
+)
+def test_propagator_extracts_empty_context(carrier: dict[str, str]) -> None:
```
willbakst: I feel like we could merge this with the text above and just name it `test_propagator_extract_context` with valid, invalid, and missing headers? We could then snapshot the bits of `Context` we need to test against to ensure it's working correctly.

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,618 @@
...
+
+    with tracer.start_as_current_span("test-span") as span:
+        current_context = trace.set_span_in_context(span)
+        carrier: dict[str, str] = {}
+        inject_context(carrier, context=current_context)
+
+        assert set(carrier.keys()) == snapshot({"traceparent"})
+
+
+def test_inject_context_with_specific_context(
```
willbakst: I feel like this should be merged with `test_propagator_injects_specific_context`?

I feel like we should just use the global methods to get the global propagator for these tests rather than constructing one directly? It'll reset after the test anyway, right?

sdks/python/tests/lilypad/_internal/test_configuration.py
```diff
@@ -229,3 +234,44 @@ def test_configure_defaults_from_zero_values() -> None:
         assert exporter.max_retry_attempts == DEFAULT_MAX_RETRY_ATTEMPTS
         _shutdown_span_processors(cast(ProxyTracerProvider, provider))
     close_cached_clients()
+
+
+@pytest.mark.parametrize(
+    "propagator_format",
+    ["tracecontext", "b3", "b3multi", "jaeger", "composite"],
```
willbakst: nit: make this `list(get_args(PropagatorFormat))`?

sdks/python/src/lilypad/_internal/otel/http/_base_instrumentor.py
koxudaxi: I see. I updated it

sdks/python/src/lilypad/_internal/otel/http/_base_instrumentor.py
```diff
@@ -0,0 +1,179 @@
...
+        """Return the log message to display after successful instrumentation."""
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def uninstrumented_log_message(self) -> str:
+        """Return the log message to display after successful uninstrumentation."""
+        raise NotImplementedError
+
+    def configure(self, **kwargs: Any) -> None:
```
willbakst: Why are we using `**kwargs: Any` here instead of specifically typing the `propagator` arg?

sdks/python/src/lilypad/_internal/otel/http/_base_instrumentor.py
```diff
@@ -0,0 +1,179 @@
...
+            propagator: Optional custom ContextPropagator to use for context injection.
+                       If not provided, the global propagator will be used.
+            **kwargs: Additional client-specific configuration options.
+
+        Subclasses can override this method to handle client-specific configuration
+        options. This method is called during _instrument before wrapping methods.
+        """
+        self._propagator = kwargs.get("propagator")
+
+    def _instrument(self, **kwargs: Any) -> None:
```
willbakst: Same thing here what is the value of `**kwargs` if we only use it for configure?

sdks/python/tests/lilypad/_internal/otel/http/test_aiohttp.py
```diff
@@ -0,0 +1,174 @@
...
+    with tracer.start_as_current_span("test-span"):
+        async with aiohttp.ClientSession() as session:
+            async with session.get(
+                url,
+                headers=MockDictHeaders(),
+            ) as response:
+                assert response.status == 200
+
+    normalized = normalize_headers(captured_headers)
+    assert "Traceparent" in normalized
```
willbakst: Using raw strings here seems brittle? Should we snapshot test `normalized` and `captured_headers`? Also maybe make it `normalized_headers` as a variable name to match?

sdks/python/tests/lilypad/_internal/conftest.py
willbakst: I'm not seeing this change? Is it be cause something at this level needs it?

sdks/python/src/lilypad/_internal/otel/http/_base_instrumentor.py
```diff
@@ -0,0 +1,179 @@
...
+
+            wrapt.wrap_function_wrapper(
+                spec.target_cls,
+                spec.method_name,
+                wrapper,
+            )
+
+        self.logger.debug(self.instrumented_log_message)
+
+    def _uninstrument(self, **kwargs: Any) -> None:
```
willbakst, line 112: What are the `**kwargs` for? We don't even use them.

sdks/python/tests/lilypad/_internal/otel/http/test_requests.py
```diff
@@ -0,0 +1,139 @@
...
+        response = session.get(url, headers=MockDictHeaders())
+
+    assert response.status_code == 200
+
+    normalized = normalize_headers(captured_headers)
+    assert normalized["Traceparent"] == "<traceparent>"
+    assert captured_headers["X-Custom-Header"] == "value"
+
+
+def test_requests_with_custom_propagator(
```
willbakst: Why is requests the only one with this test? If we don't need to test this across all of the integrations, shouldn't we just test the base class directly?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,615 @@
...
+
+
+@pytest.fixture
+def tracer(tracer_provider: TracerProvider) -> trace.Tracer:
+    """Get a tracer from the test provider."""
+    return trace.get_tracer(__name__)
+
+
+@pytest.mark.parametrize(
+    ("propagator_env", "expected_headers"),
```
willbakst: nit: parameter should be `propagator_format` to match?

sdks/python/tests/lilypad/_internal/test_configuration.py
```diff
@@ -229,3 +235,44 @@ def test_configure_defaults_from_zero_values() -> None:
         assert exporter.max_retry_attempts == DEFAULT_MAX_RETRY_ATTEMPTS
         _shutdown_span_processors(cast(ProxyTracerProvider, provider))
     close_cached_clients()
+
+
+@pytest.mark.parametrize(
+    "propagator_format",
+    list(get_args(PropagatorFormat)),
```
willbakst: oh, wait do we have a `PROPAGATOR_FORMATS` constant somewhere?

sdks/python/tests/lilypad/_internal/conftest.py
koxudaxi: It had been reverted, so I moved it again.

sdks/python/src/lilypad/_internal/otel/http/_base_instrumentor.py
koxudaxi: I remember now. Since it inherits from `opentelemetry.instrumentation.instrumentor.BaseInstrumentor`, we cannot remove the `kwargs`.

sdks/python/src/lilypad/_internal/otel/http/_base_instrumentor.py
koxudaxi: same here.
> I remember now. Since it inherits from `opentelemetry.instrumentation.instrumentor.BaseInstrumentor`, we cannot remove the `kwargs`.

sdks/python/tests/lilypad/_internal/otel/http/test_requests.py
koxudaxi: OK, I added the test for base class.

sdks/python/src/lilypad/_internal/otel/http/_base_instrumentor.py
willbakst: Got it. Adding `propagator` as a separated out arg solves this for me.

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
```diff
@@ -0,0 +1,614 @@
...
+    return trace.get_tracer(__name__)
+
+
+@pytest.mark.parametrize(
+    ("propagator_format", "expected_headers"),
+    PROPAGATOR_FORMAT_CONFIGS,
+    ids=["default"] + PROPAGATOR_FORMATS,
+)
+def test_propagator_formats(
+    propagator_format: str | None,
```
willbakst: nit: technically should have type

`propagator_format: PropagatorFormat | None`

sdks/python/tests/lilypad/_internal/otel/http/test_aiohttp.py
```diff
@@ -0,0 +1,187 @@
...
+
+    tracer = trace.get_tracer(__name__)
+
+    with tracer.start_as_current_span("test-span"):
+        async with aiohttp.ClientSession() as session:
+            async with session.get(capture.urls[0]) as response:
+                assert response.status == 200
+
+    normalized_before = normalize_headers(capture.captured_headers[0])
+    assert "Traceparent" not in normalized_before
```
willbakst, line 73: We're still using the raw string here. Right now this is what we're testing:

```python
assert "Traceparent" not in normalized_before

assert normalized_instrumented == snapshot({...})

assert "Traceparent" not in normalized_after
```

Instead, I think what we really want to test is something like this:

```python
assert normalized_before == normalized_after
assert "Traceparent" in normalized_instrumented
assert normalized_instrumented.pop("Traceparent") == "<traceparent>"
assert normalized_instrumented == normalized_before
```

The main thing the current test is missing that I'm realizing is testing that the only thing the instrumentation has done is insert the `"Traceparent"` key while ensuring everything else is the same.

Let's make this update here as well as for the other instrumentations?

sdks/python/tests/lilypad/_internal/otel/test_propagation.py
koxudaxi: I fixed remaining `str`

willbakst (approved):

willbakst:
### Merge activity

* **Oct 17, 2:13 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/lilypad/706).

PR #707, opened 2025-10-07
Implement with lilypad.session context manager for grouping traces

koxudaxi:
### TL;DR

Added session context management to group related traces together across service boundaries.

### What changed?

- Added a new session context management system that allows grouping related traces together
- Implemented `session()` context manager and `current_session()` function to create and access session contexts
- Added automatic propagation of session IDs via HTTP headers (`Lilypad-Session-Id`)
- Enhanced spans to automatically capture session information as attributes
- Integrated session extraction with the existing context propagation system

### How to test?

```python
# Create a session to group related traces
with lilypad.session(id="user-123", attributes={"role": "admin"}):
    # All spans created in this context will be associated with this session
    with lilypad.span("operation"):
        # Make HTTP requests - session ID will be propagated automatically
        response = requests.get("https://api.example.com")
        
    # Check current session
    session_ctx = lilypad.current_session()
    print(session_ctx.id)  # "user-123"
```

### Why make this change?

This feature enables better correlation of traces across service boundaries by providing a higher-level grouping mechanism than trace IDs alone. Sessions allow developers to logically group related operations (potentially spanning multiple traces) that belong to the same user interaction or business transaction, making it easier to analyze and debug complex distributed systems.


koxudaxi:
* **#708** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/708?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#707** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/707?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/707?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#706** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/706?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v1`


<h2></h2>

sdks/python/src/lilypad/_internal/otel/propagation.py
```diff
@@ -164,6 +173,9 @@ def inject_context(
 ) -> None:
     """Inject current OTEL context into carrier headers using the global propagator.
 
+    This function also injects session context if one is active, adding the
+    Lilypad-Session-Id header to the carrier.
```
willbakst: nit: this should be `SESSION_HEADER_NAME header to the carrier` since we use the variable here?

sdks/python/src/lilypad/_internal/session.py
```diff
@@ -0,0 +1,132 @@
...
+SESSION_CONTEXT: ContextVar[SessionContext | None] = ContextVar(
+    "SESSION_CONTEXT", default=None
+)
+
+
+@dataclass
+class SessionContext:
+    """Represents a session context for grouping related traces."""
+
+    id: str
```
willbakst: nit: give the fields docstrings

sdks/python/src/lilypad/_internal/session.py
```diff
@@ -0,0 +1,132 @@
...
+    attributes: dict[str, Jsonable] | None = None
+
+    def __post_init__(self) -> None:
```
willbakst: nit: missing docstring

sdks/python/src/lilypad/_internal/session.py
```diff
@@ -0,0 +1,132 @@
...
+    Raises:
+        ValueError: If id is empty or attributes are not JSON-serializable.
+
+    Example:
+        ```python
+        with lilypad.session(id="user-123"):
+            result = requests.get("https://api.example.com")
+        ```
+
+        Nested sessions override parent session:
```
willbakst: Either merge these into a single example or have two distinct `Example: ...` sections for each example

sdks/python/src/lilypad/_internal/session.py
```diff
@@ -0,0 +1,132 @@
...
+            # Session ID: 1
+            with lilypad.session(id="2"):
+                # Session ID: 2
+            # Session ID: 1
+        ```
+    """
+    session_ctx = SessionContext(id=id, attributes=attributes)
+    token: Token[SessionContext | None] = SESSION_CONTEXT.set(session_ctx)
+    try:
+        yield
```
willbakst: Shouldn't we yield the `SessionContext`?

sdks/python/src/lilypad/_internal/session.py
```diff
@@ -0,0 +1,132 @@
...
+    attributes: dict[str, Jsonable] | None = None
+
+    def __post_init__(self) -> None:
+        if not self.id or not isinstance(self.id, str):
+            raise ValueError("Session ID must be a non-empty string")
+
+
+@contextmanager
+def session(
+    *, id: str, attributes: dict[str, Jsonable] | None = None
```
willbakst: We should allow for `id: str | None = None` where if a user does not provide an ID we will generate something for them under the hood.

sdks/python/tests/lilypad/_internal/test_context.py
```diff
@@ -198,3 +201,109 @@ def test_propagated_context_with_empty_carrier() -> None:
...
+            assert lilypad.current_session() is not None
+            raise RuntimeError("test exception")
+    except RuntimeError:
+        pass
+
+    assert lilypad.current_session() is None
+
+
+def test_propagated_context_with_b3_propagator(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Verify B3 propagator works with session extraction (regression test)."""
```
willbakst: What is this a regression test of? Why not test this against various propagators?

sdks/python/tests/lilypad/_internal/test_span.py
willbakst, line 1: We should make sure we are using our span implementation everywhere (e.g. trace and version) so that those also have session support.

sdks/python/tests/lilypad/_internal/otel/http/header_patterns.py
```diff
@@ -0,0 +1,21 @@
+"""Canonical header patterns for HTTP instrumentation testing."""
```
willbakst: I feel like I've seen these somewhere else already? Is there somewhere that is not using these refactored versions?

sdks/python/src/lilypad/_internal/session.py
```diff
@@ -0,0 +1,150 @@
...
+    attributes: dict[str, Jsonable] | None = None
+
+    def __post_init__(self) -> None:
+        """Validate session context after initialization.
+
+        Raises:
+            ValueError: If session ID is empty or not a string.
+        """
+        if not self.id or not isinstance(self.id, str):
+            raise ValueError("Session ID must be a non-empty string")
```
willbakst: Why raise a `ValueError` here? I think it would be better to handle the `self.id is None` case here and generate a valid ID (rather than pushing that off to the consumer).

sdks/python/src/lilypad/_internal/session.py
willbakst: I see you added the `Attributes: ...` section, but for fields let's use the per-field docstring format we've used elsewhere. That's what we're going to be pulling when we generate our API docs.

sdks/python/src/lilypad/_internal/session.py
```diff
@@ -0,0 +1,150 @@
...
+    within the session context and are included in outgoing HTTP requests.
+
+    Args:
+        id: Unique identifier for the session. If not provided, a UUID will be
+            automatically generated.
+        attributes: Optional dictionary of JSON-serializable attributes to
+            attach to the session.
+
+    Raises:
+        ValueError: If id is an empty string or attributes are not JSON-serializable.
```
willbakst: This does not actually raise anything. If the user betrays the type hinting, then there may be an error or undefined behavior. But following the type hinting won't result in such behavior.

This function itself has no `raise` clause anywhere.

sdks/python/src/lilypad/_internal/session.py
koxudaxi: Oh, OK. I made mistake. I fix it.

sdks/python/tests/lilypad/_internal/otel/http/header_patterns.py
koxudaxi: OK, I just clean up duplicated code.

sdks/python/src/lilypad/_internal/session.py
```diff
@@ -0,0 +1,147 @@
...
+    def __post_init__(self) -> None:
+        """Initialize session context, generating ID if needed.
+
+        Raises:
+            ValueError: If session ID is an empty string or not a string type.
+        """
+        if self.id is None:
+            self.id = str(uuid.uuid4())
+        elif not isinstance(self.id, str) or not self.id:
+        if self.id is None or not isinstance(self.id, str) or not self.id:
```
willbakst: Something weird happened here with Graphite that added this but didn't delete the existing code...

sdks/python/src/lilypad/_internal/session.py
```diff
@@ -0,0 +1,147 @@
...
+SESSION_CONTEXT: ContextVar[SessionContext | None] = ContextVar(
+    "SESSION_CONTEXT", default=None
+)
+
+
+@dataclass
+class SessionContext:
+    """Represents a session context for grouping related traces."""
+
+    id: str | None = None
```
willbakst: I think `id: str` with no default is correct here since `SessionContext` should always have an `id`.

You would get rid of the `__post_init__` and have something like this:

```python
id: str = field(default=str(uuid.uuid4()))
```

sdks/python/src/lilypad/_internal/session.py
```diff
@@ -0,0 +1,147 @@
...
+    """Unique identifier for the session. Auto-generated if not provided."""
+
+    attributes: dict[str, Jsonable] | None = None
+    """Optional JSON-serializable metadata associated with the session."""
+
+    def __post_init__(self) -> None:
+        """Initialize session context, generating ID if needed.
+
+        Raises:
+            ValueError: If session ID is an empty string or not a string type.
```
willbakst: Still seeing this.

In line with my other comment, we should just make `id` a field with a default value of `str(uuid.uuid4())`

sdks/python/src/lilypad/_internal/spans.py
```diff
@@ -61,6 +62,15 @@ def __enter__(self) -> Span:
         else:
             self._is_noop = False
             self._span.set_attribute("lilypad.type", "trace")
+
+            session_ctx = current_session()
+            if session_ctx is not None and session_ctx.id is not None:
```
willbakst: Given my other comment, `if session_ctx is not None` then `session_ctx.id` will be populated.

sdks/python/src/lilypad/_internal/session.py
willbakst: Given other comments, this should be deleted entirely since `id` will have a default

willbakst (approved):

willbakst:
### Merge activity

* **Oct 17, 2:17 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/lilypad/707).

PR #708, opened 2025-10-09
Update existing OpenAI and Anthropic instrumentation to match 1.37.0

koxudaxi:
### TL;DR

Updated OpenTelemetry instrumentation to use the new GenAI semantic conventions format for LLM telemetry.

### What changed?

- Refactored the OpenAI and Anthropic instrumentors to use the new OpenTelemetry GenAI semantic conventions
- Replaced the event-based telemetry approach with structured attributes using the new `gen_ai.input.messages`, `gen_ai.output.messages`, and `gen_ai.system_instructions` fields
- Added support for structured message parts (text, tool calls, tool responses) in a format compatible with the latest OTel spec
- Updated streaming response handling to collect complete messages before sending telemetry
- Improved JSON handling for tool calls to better support both string fragments and complete objects

### How to test?

1. Run the existing tests for OpenAI and Anthropic instrumentation
2. Verify that spans now include the new attributes format with structured messages
3. Test with both streaming and non-streaming responses
4. Test with tool calls to ensure proper formatting of arguments

### Why make this change?

This change aligns our OpenTelemetry instrumentation with the latest GenAI semantic conventions, providing more structured and consistent telemetry data. The new format makes it easier to analyze LLM interactions in observability platforms and provides better support for complex message structures including tool calls, multimodal content, and streaming responses.


koxudaxi:
* **#708** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/708?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/708?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#707** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/707?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#706** <a href="https://app.graphite.dev/github/pr/Mirascope/lilypad/706?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v1`


<h2></h2>

sdks/python/src/lilypad/_internal/otel/anthropic.py
```diff
@@ -181,50 +116,216 @@ def _get_request_attributes(
...
+                            if not isinstance(block, dict):
+                                if block.type == "text":
+                                    user_parts.append(
+                                        InputTextPart(
+                                            type="text",
+                                            content=block.text,
+                                        )
+                                    )
+                            else:
+                                if block["type"] == "text":
```
koxudaxi: @willbakst 
I feel the nesting is getting a bit too deep in the method.
Do you have any recommended approach for handling this?

sdks/python/src/lilypad/_internal/otel/stream_wrappers.py
```diff
@@ -115,22 +107,36 @@ def append_tool_call(
...
-            index=self.index,
-            message=message,
+            parts.append(
+                OutputTextPart(
+                    type="text",
+                    content=self.content,
+                )
+            )
+
+        for buffer in self.tool_calls_buffers:
```
willbakst: We don't actually know the order with this current implementation. We'll need to update the implementation to maintain `content` as a list of `str` or `ToolCallBuffer`, and then we should add parts here according to the order in which they were collected so it's preserved.

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -289,13 +397,29 @@ def dump(self) -> dict[str, AttributeValue]:
...
+                "input_messages",
+                "GEN_AI_PROVIDER_NAME",
+            ]:
                 continue
             value = getattr(self, field.name)
             if value is None:
                 continue
             attributes[getattr(gen_ai_attributes, field.name)] = value
 
+        if self.GEN_AI_PROVIDER_NAME:
```
willbakst: This implementation doesn't utilize the existing dynamic attribute matching of the fields.

If `gen_ai_attributes` has a field `GEN_AI_PROVIDER_NAME`, it should work with the existing mapping. Are there similar attributes for `system_instructions` / `input_messages`?

Also, I'm open to a different implementation approach. For example, I'm more than happy to implement direct setting for all fields if we think that's cleaner since it would only be in this one place.

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -312,169 +436,39 @@ class GenAIResponseAttributes:
...
-
-    content: Any | None = None
-    """The content of the tool message."""
+            value = getattr(self, field.name)
+            if value is not None:
+                attributes[getattr(gen_ai_attributes, field.name)] = value
 
-    id: str | None = None
-    """The identifier for the tool call."""
+        if self.output_messages:
```
willbakst: Same comment here.

sdks/python/tests/lilypad/_internal/otel/test_openai.py
willbakst, line 1: How can this have +972 lines but `test_anthropic.py` only has `451`? What is it about the OpenAI implementation that needs so much more testing? Is it that many more types to handle? I feel like it should be somewhat the same?

sdks/python/src/lilypad/_internal/otel/anthropic.py
```diff
@@ -181,50 +116,216 @@ def _get_request_attributes(
...
+                                            ToolCallResponsePart(
+                                                type="tool_call_response",
+                                                id=tool_use_id,
+                                                result=tool_content,
+                                            )
+                                        )
+                                    elif isinstance(tool_content, list):
+                                        for content_block in tool_content:
+                                            if (
+                                                isinstance(content_block, dict)
```
willbakst: What could it be if not a `dict`? There doesn't seem to be any additional handling of the else case?

sdks/python/src/lilypad/_internal/otel/anthropic.py
willbakst: The `_process_messages` method is no longer useful as a static method because it's only used by the `_get_request_attributes` method that consumes it.

I think we should refactor the `_process_messages` method in each provider into private utilities. Even just this will be a good first step wrt. the indentation problem since it will lose the class level indentation immediately. I would then rename `_get_request_attributes` to `_process_request` instead.

I would then try to refactor some of the block handling that looks similar across user input, assistant input, and assistant output messages. I might have private utility methods for handling each time (like before with e.g. `_process_assistant_message`.

After doing this I would see how I feel about the indentation / implementation and see whether or not it would need any additional cleanup.

sdks/python/src/lilypad/_internal/otel/anthropic.py
```diff
@@ -181,50 +116,216 @@ def _get_request_attributes(
...
+
+        for block in response.content:
+            if block.type == "text":
+                parts.append(
+                    OutputTextPart(
+                        type="text",
+                        content=block.text,
+                    )
+                )
+            elif block.type == "tool_use":
```
willbakst, line 245: What is the assistant has an output part? I think if you use a provider's native tool, it will call it on it's server and include the output in the assistant message?

sdks/python/src/lilypad/_internal/otel/anthropic.py
koxudaxi: OK, I just refactored.

sdks/python/src/lilypad/_internal/otel/anthropic.py
```diff
@@ -84,75 +112,226 @@ class _AnthropicKwargs(BaseKwargs, total=False):
...
+) -> list[SystemInstruction]:
+    """Process system instructions from kwargs.
+
+    Args:
+        system_content: System content from Anthropic kwargs.
+
+    Returns:
+        List of SystemInstruction objects.
+    """
+    system_instructions: list[SystemInstruction] = []
```
willbakst: nit: we can simplify this slightly:

```python
if isinstance(system_content, str):
    return [SystemInstruction(type="text", content=system_content)]

return [
    SystemInstruction(type="text", content=block.get("text", ""))
    for block in system_content
]
```

sdks/python/src/lilypad/_internal/otel/anthropic.py
```diff
@@ -84,75 +112,226 @@ class _AnthropicKwargs(BaseKwargs, total=False):
...
+
+    Args:
+        content: Assistant message content from Anthropic.
+
+    Returns:
+        List of InputMessagePart objects.
+    """
+    parts: list[InputMessagePart] = []
+
     if isinstance(content, str):
```
willbakst, line 103: We can remove another level of indentation here if we do this:

```python
if isinstance(content, str):
    return [InputTextPart(type="text", content=content)]

parts: list[InputMessagePart] = []
for block in content:
   ...
```

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -288,13 +401,62 @@ def dump(self) -> dict[str, AttributeValue]:
...
+                self.GEN_AI_OPENAI_REQUEST_SEED
+            )
+        if self.GEN_AI_OPENAI_REQUEST_SERVICE_TIER:
+            attributes[gen_ai_attributes.GEN_AI_OPENAI_REQUEST_SERVICE_TIER] = (
+                self.GEN_AI_OPENAI_REQUEST_SERVICE_TIER
+            )
+
+        attributes[_GEN_AI_PROVIDER_NAME] = self.GEN_AI_PROVIDER_NAME
+
+        if self.system_instructions:
```
willbakst: We still aren't matching it here. Some keys are all caps e.g. `GEN_AI_REQUEST_MAX_TOKENS` but then we have `self.system_instructions` which maps to `attributes[_GEN_AI_SYSTEM_INSTRUCTIONS]`, which I would guess would still conform to the existing system from before.

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -288,13 +401,62 @@ def dump(self) -> dict[str, AttributeValue]:
...
+            attributes[_GEN_AI_SYSTEM_INSTRUCTIONS] = json_dumps(
```
willbakst: Yeah this is worse than before. We should use the existing system + `serialize_attribute_value` (which you removed) since that's cleaner and achieves the same goal and makes adding new fields easier.

sdks/python/tests/lilypad/_internal/otel/test_anthropic.py
```diff
@@ -626,3 +458,330 @@ def test_anthropic_instrumentation_idempotent() -> None:
     second_create = client.messages.create
 
     assert second_create is first_create
+
+
+@pytest.mark.vcr()
+def test_anthropic_system_list_content(
```
willbakst, line 569: I want to comb through these tests to see how many tests we need for individual content types.

It might be easier to review them as separate tests, but with these full snapshots it isn't.

I want if we should be snapshotting just the `gen_ai.input.messages` and `gen_Ai.output.messages` fields?

Something like:

```python
span_data = extract_span_data(spans[0])

assert span_data["attributes"]["gen_ai.input.messages"] == snapshot([...])

assert span_data["attributes"]["gen_ai.output.messages"] == snapshot([...])
```

sdks/python/tests/lilypad/_internal/otel/test_openai.py
```diff
@@ -70,39 +71,19 @@ def test_openai_simple(client: OpenAI, span_exporter: InMemorySpanExporter) -> N
                 "server.address": "api.openai.com",
                 "gen_ai.request.stop_sequences": ("Hi!",),
                 "gen_ai.system": "openai",
+                "gen_ai.provider.name": "openai",
```
willbakst, line 91: Same comment here re: simplifying the snapshot testing for these tests to only the specific thing we care about that we're specifically testing (i.e. the message handling)?

Beyond message handling, the most basic test should cover all of the other attributes, right?

sdks/python/src/lilypad/_internal/otel/types.py
willbakst: If you're not sure what I mean here, please let me know so we can sync on this bit.

koxudaxi: I modified custom attributes that don't exist in the OpenTelemetry standard to use string literals instead of constants.

koxudaxi: I've read all your comments and made various fixes. Could you please confirm whether the current code incorporates all the changes you pointed out?

sdks/python/src/lilypad/_internal/otel/types.py
koxudaxi: I just reverted `serialize_attribute_value`. But, `serialize_attribute_value ` do only `json_dumps`. Do you expect this function? Or, Should I add another logic?

sdks/python/src/lilypad/_internal/otel/anthropic.py
willbakst: I'm not seeing any comment or change here?

sdks/python/src/lilypad/_internal/otel/anthropic.py
```diff
@@ -84,75 +112,213 @@ class _AnthropicKwargs(BaseKwargs, total=False):
...
-                            arguments=json_dumps(block.input),
-                        ),
+                user_parts.append(
+                    InputTextPart(
+                        type="text",
+                        content=block.text,
                     )
                 )
         else:
             if block["type"] == "text":
```
willbakst: Why do we use `.get("type")` above but `["type"]` here?

sdks/python/src/lilypad/_internal/otel/anthropic.py
```diff
@@ -84,75 +112,213 @@ class _AnthropicKwargs(BaseKwargs, total=False):
...
+                role="user",
+                parts=[InputTextPart(type="text", content=content)],
+            )
+        ]
+
+    messages: list[InputMessage] = []
+    user_parts: list[InputMessagePart] = []
     for block in content:
         if not isinstance(block, dict):
             if block.type == "text":
```
willbakst: This can only be text? What about all of the possible tool blocks?

sdks/python/src/lilypad/_internal/otel/anthropic.py
```diff
@@ -84,75 +112,213 @@ class _AnthropicKwargs(BaseKwargs, total=False):
...
+                        parts=tool_parts,
                     )
                 )
-    return text, tool_calls
+            else:
+                block_type = block.get("type", "")
+                if block_type:
+                    user_parts.append(InputOtherPart(type=block_type, data=block))
+
+    if user_parts:
```
willbakst: It seems weird that we may return an empty list when it should most likely be a user message with an empty parts list.

Let's update this to just

```python
return InputMessage(role="user", parts=user_parts)
```

and update the return type from `list[InputMessage]` to just `InputMessage` since we'll only ever emit a single message in this case

sdks/python/src/lilypad/_internal/otel/openai.py
```diff
@@ -144,8 +149,14 @@ def _get_request_attributes(
         service_tier = kwargs.get("service_tier")
         if service_tier == "auto":
             service_tier = None
+
+        system_instructions, input_messages = _OpenAIInstrumentor._process_messages(
```
willbakst: `_process_messages` should just be a private method of this module, not the class

sdks/python/src/lilypad/_internal/otel/openai.py
```diff
@@ -157,86 +168,224 @@ def _get_request_attributes(
...
+        input_messages: list[InputMessage] = []
+
         for message in kwargs.get("messages", []):
             match message["role"]:
                 case "system":
-                    message_events.append(
-                        SystemMessageEvent(
-                            system=_OPENAI_SYSTEM,
-                            content=message["content"],
+                    content = message.get("content", "")
```
willbakst: We should refactor this like you did for Anthropic into helpers for processing user vs. assistant vs. system messages.

sdks/python/src/lilypad/_internal/otel/openai.py
```diff
@@ -157,86 +168,224 @@ def _get_request_attributes(
...
+                    if isinstance(content, str):
```
willbakst: Same comment here as for Anthropic on the refactored method to remove some indentation with early returns.

sdks/python/src/lilypad/_internal/otel/openai.py
```diff
@@ -157,86 +168,224 @@ def _get_request_attributes(
...
+                                    )
+                        else:
+                            parts.append(
+                                ToolCallResponsePart(
+                                    type="tool_call_response",
+                                    id=tool_call_id,
+                                    result=str(content),
+                                )
+                            )
+                    input_messages.append(InputMessage(role="tool", parts=parts))
```
willbakst: Can input messages have `role="tool"`? Does that conform to the spec of `InputMessage`? I thought they just had user messages with tool call response parts?

sdks/python/src/lilypad/_internal/otel/openai.py
```diff
@@ -157,86 +168,224 @@ def _get_request_attributes(
...
+                                parts.append(OutputOtherPart(type=part_type, data=part))
+                else:
+                    parts.append(
+                        OutputTextPart(
+                            type="text",
+                            content=str(content),
+                        )
+                    )
+
+            tool_calls = _get_tool_call_params(choice.message)
```
willbakst: Why not refactor all of these to helper methods given how long they are?

sdks/python/src/lilypad/_internal/otel/stream_wrappers.py
```diff
@@ -74,63 +66,78 @@ class _ChoiceBuffer:
...
-            system=self.system,
-            index=self.index,
-            message=message,
+    def to_output_message(self) -> OutputMessage:
+        """Returns the accumulated choice data as an OutputMessage, preserving order."""
+        output_parts: list[OutputMessagePart] = []
+
+        for part in self.parts:
+            if isinstance(part, str):
+                # Text content
```
willbakst: nit: unnecessary comment

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -288,13 +415,64 @@ def dump(self) -> dict[str, AttributeValue]:
...
+            )
+        if self.GEN_AI_OPENAI_REQUEST_SERVICE_TIER:
+            attributes[gen_ai_attributes.GEN_AI_OPENAI_REQUEST_SERVICE_TIER] = (
+                self.GEN_AI_OPENAI_REQUEST_SERVICE_TIER
+            )
+
+        attributes["gen_ai.provider.name"] = self.GEN_AI_PROVIDER_NAME
+
+        if self.GEN_AI_SYSTEM_INSTRUCTIONS:
+            attributes["gen_ai.system_instructions"] = serialize_attribute_value(
```
willbakst: Do these not have corresponding `gen_ai_attributes` variables like the others?

The previous code should work here to ensure we serialize all keys and don't miss any:

```python
for field in fields(self):
    if field.name in ["SERVER_ADDRESS", "SERVER_PORT"]:
        continue
    value = getattr(self, field.name)
    if value is None:
        continue
    attributes[getattr(gen_ai_attributes, field.name)] = value
```

In this code for any attributes that don't have a variable in `gen_ai_attributes` we insert them first and then skip them in this loop. Otherwise, we gete the key name by dynamically accessing the variables, which has a matching name in the `gen_ai_attributes` module as our named fields (which is why they are all caps).

No need to manually write all of these out given we lose the benefit of being to add new fields easily in a way that includes them in the dump.

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -312,22 +490,51 @@ class GenAIResponseAttributes:
...
+    """The output messages from the LLM response."""
+
     def dump(self) -> dict[str, AttributeValue]:
         """Returns the attributes as a dictionary suitable for OpenTelemetry spans."""
-        return {
-            getattr(gen_ai_attributes, field.name): value
-            for field in fields(self)
-            if (value := getattr(self, field.name)) is not None
-        }
+        attributes: dict[str, AttributeValue] = {}
```
willbakst, line 464: Same comment here.

sdks/python/src/lilypad/_internal/otel/types.py
willbakst: I've commented on this in another comment more specifically, so I will resolve this in favor of that discussion.

sdks/python/src/lilypad/_internal/otel/types.py
willbakst: Let's sync on this and resolve as part of the other thread.

sdks/python/src/lilypad/_internal/otel/types.py
willbakst: We should keep and use the original `serialize_attribute_value` method since that properly serializes `Any` into `AttributeValue | None`. Simply calling `json_dumps` is actually incorrect as pointed out by Dandelion in a previous review of this code.

sdks/python/tests/lilypad/_internal/otel/test_anthropic.py
```diff
@@ -626,3 +405,254 @@ def test_anthropic_instrumentation_idempotent() -> None:
...
+        max_tokens=50,
+    )
+
+    assert isinstance(response, Message)
+    spans = span_exporter.get_finished_spans()
+    assert len(spans) == 1
+
+    span_data = extract_span_data(spans[0])
+
+    assert span_data["attributes"]["gen_ai.input.messages"] == snapshot(
```
willbakst, line 661: We should JSON load these values so the snapshot can be formatted more cleanly. that will make these significantly easier to read/review.

sdks/python/tests/lilypad/_internal/otel/test_openai.py
```diff
@@ -61,7 +62,21 @@ def test_openai_simple(client: OpenAI, span_exporter: InMemorySpanExporter) -> N
 
     spans = span_exporter.get_finished_spans()
     assert len(spans) == 1
-    assert extract_span_data(spans[0]) == snapshot(
+
+    span_data = extract_span_data(spans[0])
+    attrs = span_data["attributes"]
+
+    assert attrs.pop("gen_ai.system_instructions") == snapshot(
+        '[{"type":"text","content":"You greet people"}]'
```
willbakst: ditto across the board for all of these.

sdks/python/src/lilypad/_internal/otel/anthropic.py
koxudaxi: In the code above within the function, it returns a list containing both the user input message and the Tool part, if present. How can I make the function return a single Message?

koxudaxi: OK, I understand your comment after I read other your comments.

sdks/python/src/lilypad/_internal/otel/anthropic.py
```diff
@@ -84,75 +88,153 @@ class _AnthropicKwargs(BaseKwargs, total=False):
...
-                            arguments=json_dumps(block["input"]),
-                        ),
+            if block.get("type") == "text":
+                user_parts.append(
+                    InputTextPart(
+                        type="text",
+                        content=block.get("text", ""),
+                    )
+                )
+            elif block["type"] == "tool_result":
```
willbakst: This is inconsistent still. Just do `block_type = block.get("type")` at the top. Also make this change in the earlier section that currently uses the walrus to assign. It'll be more clear in that case without the walrus

sdks/python/src/lilypad/_internal/otel/anthropic.py
```diff
@@ -84,75 +88,153 @@ class _AnthropicKwargs(BaseKwargs, total=False):
...
+                )
+            elif block_type:
+                parts.append(InputOtherPart(type=block_type, data=block))
+
+    return parts
+
+
+def _process_user_message_content(
+    content: str | Iterable[ContentBlock | ContentBlockParam],
+) -> InputMessage:
```
willbakst: We should merge `_process_assistant_message_content` and this `_process_user_message_content` into a single method `_process_content` since they are equivalent and take the same input types? Then we can have one function that handles all the content types rather than duplicating like this.

Then below we can create the `InputMessage` types from the `list[InputMessagePart]` return.

sdks/python/src/lilypad/_internal/otel/openai.py
```diff
@@ -115,6 +145,161 @@ def _get_tool_call_params(
...
+    content: str | Iterable[ChatCompletionContentPartParam],
+) -> list[InputMessagePart]:
+    """Returns input message parts from OpenAI user message content."""
+    if isinstance(content, str):
+        return [InputTextPart(type="text", content=content)]
+
+    return _process_content_parts(content)
+
+
+def _process_assistant_message_content(
```
willbakst: I think we should merge the handling of content into a single `_process_content` method that returns `list[InputMessagePart]`.

Then you can use `_process_content` in the match below to get the pieces of each method. For example, you could call `_get_tool_call_params` there and extend parts there since that's really the only distinct functionality (while everything else is shared)

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -1,34 +1,66 @@
...
+def _is_str_list(value: list) -> TypeIs[list[str]]:
+    """Returns True if value is a non-empty list of strings."""
+    return (
+        bool(value)
+        and isinstance(value[0], str)
+        and all(isinstance(item, str) for item in value)
+    )
+
+
+def serialize_attribute_value(value: SerializableValue) -> AttributeValue:
```
willbakst: please revert this back to the original, which we had already tested? What was wrong with that one? To me it seems more robust and handles things like `list[float]` etc.?

sdks/python/src/lilypad/_internal/otel/types.py
```diff
@@ -274,26 +408,68 @@ class GenAIRequestAttributes:
...
         attributes: dict[str, AttributeValue] = {}
 
         if self.SERVER_ADDRESS:
             attributes[server_attributes.SERVER_ADDRESS] = self.SERVER_ADDRESS
-        if self.SERVER_PORT and self.SERVER_PORT > 0 and self.SERVER_PORT != 443:
-            attributes[server_attributes.SERVER_PORT] = (
-                self.SERVER_PORT
-            )  # pragma: no cover
+
+        if self.GEN_AI_PROVIDER_NAME:
```
willbakst: I still don't understand why you aren't including these in the dynamic functionality.

Let's sync on this so I can see the code / type hints.

willbakst (approved):

willbakst:
### Merge activity

* **Oct 17, 2:20 AM UTC**: @willbakst merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Mirascope/lilypad/708).

PR #722, opened 2025-12-01
fix: Allow None for function_uuid and add clear error for annotations without versioning

alisvanni:
## Problem
When using `@lilypad.trace(versioning=False, mode="wrap")`, attempting to call `trace.annotate()` fails with a cryptic UUID validation error:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for AnnotationCreate
function_uuid
  UUID input should be a string, bytes or UUID object [type=uuid_parsing, input_value='None', input_type=str]
```

The error message shows `input_value='None'` (string) instead of Python None, which is confusing.
## Root Cause
The `_TraceBase.__init__()` declares `function_uuid: str`, but when `versioning=False`, the decorator passes `function_uuid=None` (Python None). This type mismatch causes `None` to be stringified to "None" during serialization, which then fails UUID validation in the API. Additionally, there's no documentation or error message indicating that **annotations require versioning to be enabled**.

## Solution
This PR fixes the issue by:
1. **Correcting type annotations**: Changed `function_uuid: str` to `function_uuid: str | None` in `_TraceBase.__init__()` to accurately reflect that None is a valid value
2. **Adding validation with clear error messages**: Added checks in both `_get_span_uuid()` and `_get_span_uuid_async()` that raise a clear `ValueError` when attempting to annotate a trace without a function_uuid:
```
Cannot annotate trace without function_uuid. 
Use versioning='automatic' in @trace decorator to enable annotations.
```
3. **No breaking changes**: This is purely additive - all existing working code continues to work exactly as before

## Changes Made
**File**: `src/lilypad/traces.py`
- Line 98: `function_uuid: str` → `function_uuid: str | None`
- Line 100: `self.function_uuid: str` → `self.function_uuid: str | None`
- Lines 133-137: Added validation check in `Trace._get_span_uuid()`
- Lines 219-223: Added validation check in `AsyncTrace._get_span_uuid()`

## Benefits
- **Better type safety**: Type annotations now match actual runtime behavior
- **Clear error messages**: Developers get actionable guidance instead of cryptic UUID validation errors
- **Improved DX**: Saves developers hours of debugging by making the versioning requirement explicit
- **No breaking changes**: Backward compatible with all existing code

## Testing
Tested with AWS Bedrock AgentCore integration where this bug was discovered. Annotations now work correctly with `versioning='automatic'`, and provide clear error messages when versioning is disabled.


sdks/python/src/lilypad/traces.py
```diff
@@ -94,10 +94,11 @@ class _TraceBase(Generic[_T]):
     """
     Base class for the Trace wrapper.
     """
-
-    def __init__(self, response: _T, span_id: int, function_uuid: str) -> None:
+    # Changed type annotation in _TraceBase class, added None to function_uuid
```
willbakst, line 97: This comment (and others below) are unnecessary. Let's remove them.

willbakst:
Thank you for this PR! Once the unnecessary comments are removed and this is merged I will release a new version with this fix :)

